PR #8, opened 2015-05-29
Change curly brackets to quotes (single)

af7:
Since the expression is evaluated the curly brackets would literally mean that a variable is expected with the name `top` (for example). This was breaking my app.


wwayne:
oh yes, my careless

PR #11, opened 2015-05-29
Change module definition to UMD

mgechev:
I had troubles using the module in non-browserify project because of the module definition. I believe using UMD will make it more universal.

The changes I made is in the [package.json file](https://github.com/wwayne/react-tooltip/compare/master...mgechev:master#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8) and later I run `npm run build:js` in order to make the tooltip available in `dist`.

### DIFF
--- a/dist/react-tooltip.js
+++ b/dist/react-tooltip.js
@@ -1,147 +1,152 @@
-'use strict';
+(function (global, factory) {
+  if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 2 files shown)


mgechev:
Nevermind...it requires rename of the `index` file and has troubles resolving the `classNames` dependency when used outside CommonJS environment.

wwayne:
@mgechev I don't have too much experience on UMD, so I need to check some documents first

PR #19, opened 2015-06-23
remove querySelector and use findDOMNode instead

bezreyhan:
I removed document.querySelector and used React.findDOMNode(this) instead. This allows us to cleanup the code a bit. It should also be a bit more efficient since we are not querying the DOM as much.


wwayne:
:+1:

PR #23, opened 2015-08-02
allow for custom class

conorhastings:
allows user to pass optional custom class for styling the tooltip


wwayne:
I agree we need to add custom class in the future. But we can find a better way to implement this, `customClassName` is a too long name for user to typing and we need to ensure the css from the custom class will override the default css.
Thanks for your PR, I need to think more about this later

kswope:
Needed for modal dialogs, 

example

```
.tooltip {
    z-index: 2000 !important;
}
```

and why sweat overriding classes?, just something like data-appendClass='mynewclass'

buddyp450:
Requesting this be re-evaluated. I was in need of a simple way to tell my tooltip to just "always appear in front", was looking to set the z-index easily.

PR #26, opened 2015-08-03
Add editorconfig and eslinter to project

johnamiahford:


wwayne:
ok, thanks you then it will standardize this component, I will also add eslint to npm script

PR #34, opened 2015-09-11
avoid getBoundingClientRect() to be called twice

ruffle1986:
It's unnecessary to enforce the browser to calculate the rectangle data twice.


wwayne:
:+1:

PR #36, opened 2015-09-12
also babelify and move style.js to dist

conorhastings:
screen shot 2015-09-12 at 6 58 09 pm

error in building with version 0.5.0 since there is no style.js in dist


wwayne:
thanks :+1:

PR #38, opened 2015-10-01
Allow inserting HTML raw code inside the span

alexnm:
I see a lot of cases where you need to format your text inside the tooltip and html formatting is the closest to the react way I think. This way you can specify the tooltip text as raw html: "<p>This is a <strong>tooltip</strong></p>"

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -249,7 +249,7 @@ export default class ReactTooltip extends Component {
     }
 
     return (
-      <span className={tooltipClass} data-id='tooltip'>{this.state.placeholder}</span>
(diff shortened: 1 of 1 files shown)


wwayne:
This may be a method to support html format content, but in this pull request, user will get [object] when they just provide content like `data-tip='tooltip'`.

We may need a better way to implement this(make this as an option perhaps)
Anyway, you've provide a good way to implement this, thanks

alexnm:
yes, I agree, an option would be great. If I have time I might try a more generic approach and do another pull request.

PR #54, opened 2015-11-26
New Feature: watchWindow

Sly777:
Prop Name: watchWindow
Default: false
Values: true | false

To watch changes of window size & reposition tooltip


wwayne:
I think it can be a required feature, not an option. let's merge it first.
:+1:  thanks

PR #55, opened 2015-11-27
Add support for borders

ahstro:
This adds the ability to add a one pixel border of the same color as the text by adding the `border` attribute to the JSX component, like this `<ReactTooltip border />` or `<ReactTooltip border={true} />`.
This also adds the new feature to the example page and tweaks heights accordingly.

dark-top
light-right

Also, I threw a small typo fix in there.


wwayne:
I used to think users can just use the feature `class` to achieve this, but looks it would be better to support a `border` feature because it's really complicated to implement a border...

But I'd like to support in the way of `<ReactTooltip border='1px solid #fff' />` that would be easier and flexible to use, so I will do some development before merge this PR or you can keep push your thoughts.

Anyway, thanks :+1:

ahstro:
My original thought was to do it like that as well, but I thought I'd create a basic version first, especially since you don't seem to support custom colors in any way other than custom class.

Borders around speech bubbles (specifically with the little arrow) are quite complicated, so I don't think you could support anything other than `solid` without making it insanely difficult and complex.

Let me know if you want me to keep working on this or if you want to take it from here? :)

wwayne:
Just checked, yes it's also difficult to set the border-with on the little arrow.
maybe something like `<ReactTooltip border='#fff' />` would be better I think

ahstro:
To me, it would make more sense to just go with these changes now for a minor release and then go for a major one where styling of the entire component is doable without custom CSS and `!important`, but if you want to I can start working on setting the border color like you mentioned above.

wwayne:
I prefer setting the border color, then I can release after polishing some details.
I can do it if you don't time, anyway, thanks a lot :+1:

ahstro:
Okay, go ahead.
No problem, thank you for a great component :)

ahstro:
Nice! :) Would you mind publishing the new version to NPM?

wwayne:
Just published in npm, there is a bug just being found, so I wanted to fix it before publishing.

ahstro:
Ah, I see. Thanks :)

PR #62, opened 2015-12-17
Fixes IE's lacking event constructor

mikkelc:
It seems like in IE an exception "Object doesn't support this action" is thrown, when using the hide and rebuild functions. This is due to the fact that IE didn't have a custom Event constructor before Edge.

For more information:
http://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-ev
ent-constructor-doesnt-work

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,8 +13,25 @@ export default class ReactTooltip extends Component {
    * Class method
    * @see ReactTooltip.hide() && ReactTooltup.rebuild()
    */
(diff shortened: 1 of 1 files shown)


wwayne:
@mikkelc you can use `npm test` to check code style, and `npm run deploy` before you push your PR.
Anyway, I will take look at this later, you don't need to care about the failure message come from `ci`.

mikkelc:
Oh... Thank you!

wwayne:
thanks :+1:

PR #68, opened 2016-01-09
better fix for Maximum call stack size exceeded

enjalot:
I've implemented a more thorough handling for tooltips going outside of the window. The problem occurs in both float or solid mode when updating the place of the tooltip will also put it outside the window, this can happen with items (like icons) placed in a corner of the screen.

This code will only update the place of the tooltip if updating it will not put the tooltip outside in anyway. In most cases this should still fix the positioning, but for the rare cases that it doesnt it will simply allow the tooltip to go outside of the window rather than erroring with Maximum call stack size exceeded.

I've re-enabled the functionality for solid since this should always work.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -394,32 +394,67 @@ export default class ReactTooltip extends Component {
         xPosition += parseInt(offset[key], 10)
       }
     }
-    /* When tooltip over the screen */
-    let offsetEffectX = effect === 'solid' ? 0 : (place ? offsetFromEffect[place].x : 0)
-    let offsetEffectY = effect === 'solid' ? 0 : (place ? offsetFromEffect[place].y : 0)
-    const styleLeft = this.state.x + offsetEffectX + xPosition
-    const styleTop = this.state.y + offsetEffectY + yPosition
-    const windoWidth = window.innerWidth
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -394,32 +394,67 @@ export default class ReactTooltip extends Component {
...
+    function getStyleLeft (_place) {
+      var _offsetEffectX = effect === 'solid' ? 0 : _place ? offsetFromEffect[_place].x : 0
+      return x + _offsetEffectX + xPosition
+    }
+    function getStyleTop (_place) {
+      var _offsetEffectY = effect === 'solid' ? 0 : _place ? offsetFromEffect[_place].y : 0
+      return y + _offsetEffectY + yPosition
+    }
+
+    var windoWidth = window.innerWidth
```
pavelloz: @enjalot Little typo :)

enjalot: hah i didn't notice. nice way to save a character ;)

wwayne:
Thanks :+1: 
Gonna make some small change(code style, es6...) before update npm

wwayne:
@enjalot  NPM published, v1.0.0

enjalot:
awesome! :100: thanks so much.

PR #69, opened 2016-01-10
account for effect=solid in outside checks

enjalot:
I missed the case where a solid tooltip would go out of bounds and the check will fail with `Cannot read property x of undefined`
I fixed by setting the default value of `offsetFromEffect` to be 0,0. Another option would be to remove the check for effect=="solid", but I'm not sure if that will have an adverse effect.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -380,6 +380,11 @@ export default class ReactTooltip extends Component {
         x: 10,
         y: -(tipHeight / 2)
       }
+    } else {
+      offsetFromEffect.top = {x: 0, y: 0}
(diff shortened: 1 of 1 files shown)


wwayne:
I tried some methods, finally set the initial value to {x:0, y:0} on your advice:
```
const placements = ['top', 'bottom', 'left', 'right']
placements.forEach(key => {
    offsetFromEffect[key] = {x:0, y:0}
})
```
I forgot to merge your code first before modification, so I have to close this PR.
Anyway, thanks :+1:

enjalot:
no sweat, glad it helped! :+1:

PR #72, opened 2016-01-20
clearing delayShowLoop-Timeout when component will unmount

Monterra:
else it will throw an error, that you can't setState of an unmounted component


wwayne:
Thanks :+1:

PR #73, opened 2016-01-20
Clear timeout when unmounting.

joelburget:
This caused an exception when the tooltip was removed from the dom
before it was scheduled to show up.


wwayne:
humm, two PR from different people for same problem, thanks :+1:

Issue #83, opened 2016-03-07
Calculate outside position for solid effect


andrejunges:
I think we should calculate the outside positions for `solid` effect as well..
We dont have to follow the mouse position (of course) but is there a case why we shouldn't considerate the outside positions for it?
![screen shot 2016-03-07 at 1 38 19 pm](https://cloud.githubusercontent.com/assets/5039334/13575900/e7ecc314-e469-11e5-95ff-ba89594cd7c0.png)

wwayne:
Hi @andrejunges , what do you mean by saying `calculate the outside positions`? I guess it is: for example, if the tooltip in top is out of the web page, it should be changed to bottom position automatically.

andrejunges:
exactly @wwayne

andrejunges:
was there a reason to not implement it? @wwayne

wwayne:
@andrejunges No, just because no time for this at the moment, may do some trying this weekend.

wwayne:
Thanks @andrejunges, support from 1.1.4

PR #84, opened 2016-03-10
Add logic to handle outside cases for solid tooltips #83

andrejunges:
@wwayne 
I implemented a simple logic to handle those cases I mentioned on the issue #83 
What do you think about it? any comments/suggestions?


wwayne:
@andrejunges Great, thanks for your help :+1:

PR #86, opened 2016-03-23
Fixed CurrentItem Bug

Sly777:
Fixed CurrentItem Bug


wwayne:
Looks the test failed and what is `void 0`?

PR #92, opened 2016-04-11
Update peer dependencies to support react / react-dom >= 15.0.0

cchamberlain:
This will get rid of the nag message for npm v3 users and will allow npm v2 users on React 15 to continue using react-tooltip.


wwayne:
Thanks :+1:

cchamberlain:
NP!

PR #95, opened 2016-04-21
Added ability to change event propagation mode from default "bubble" to "capture"

aweber1:
I had a scenario where the tooltip needed to be triggered on click. However, the trigger element is contained within a parent component that needed to handle the trigger click event as well. Unfortunately, the tooltip checkStatus method stopped click propagation in order to add a handler for outside click events that will close the tooltip.

I tried using a custom event to activate the tooltip on trigger click. This almost worked as the click event was propagating but unfortunately the event was bubbling _after_ the checkStatus outside handler was registered, so the tooltip was being hidden immediately.

The change I ultimately ended up making was to introduce a new prop to set the propagation mode to capture instead of bubble. This effectively disableds the e.stopPropagation call in checkStatus and changes the outside click handler to "capture" instead of "bubble", which allows the parent click event to be handled first and then displays the tooltip. And outside click events still close the tooltip. 

I was wary that changing the default bubble behavior might have unintended consequences for other people using this component, so the default event propagation mode will always be "bubble". But by setting the "eventPropagationMode" prop to "capture", you can change the default behavior.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -195,14 +195,16 @@ class ReactTooltip extends Component {
    * Used in customer event
    */
   checkStatus (e) {
-    e.stopPropagation()
+    if (this.props.eventPropagationMode === 'bubble') {
+      e.stopPropagation()
+    }
     if (this.state.show && e.currentTarget.getAttribute('currentItem') === 'true') {
       this.hideTooltip(e)
     } else {
       e.currentTarget.setAttribute('currentItem', 'true')
       /* when click other place, the tooltip should be removed */
       window.removeEventListener('click', this.bindClickListener)
-      window.addEventListener('click', this.bindClickListener, false)
+      window.addEventListener('click', this.bindClickListener, this.props.eventPropagationMode === 'capture')
 
       this.showTooltip(e)
       this.setUntargetItems(e.currentTarget)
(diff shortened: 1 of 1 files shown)


megalithic:
👍 

i had just forked this repo to deal with this same issue. thanks @aweber1

wwayne:
@aweber1 Thanks, merging, will add related doc when I have time

megalithic:
@wwayne shouldn't the docs and example be updated? at the very least the readme.md with the new prop, etc? i think this would be beneficial to others to have publicized. /cc @aweber1

JeremyColton:
I have a Label functional component:
```
   <label className={props.className} htmlFor={props.htmlFor}>

        <span>{props.title}</span>

        <a data-tip data-for={props.title} data-iscapture='false'>
            <i className={`anyclip-tooltip fa fa-fw fa-info-circle`}/>
        </a>

        <ReactTooltip id={props.title} className='extraClass' delayHide={100} effect='solid' type='dark'>
               My tooltip text
         </ReactTooltip>
    </label>
```

When the related input field is a checkbox and the `props.htmlFor` is provided, clicking the tooltip's icon also changes the state of the related checkbox. I tried adding `data-iscapture='false'` but it doesn't prevent even bubbling to the parent. Is my syntax wrong? 

Thanks.

PR #98, opened 2016-05-02
Adding 'hideOnClick' option

lukeknep:
This adds an option to hide the tooltip when the target element is clicked.

In my app, I have a tooltip on a button that is moved / hides itself after it is clicked. When the button is clicked, the button disappears, but the tooltip stays present until I mouse over something else that has a tooltip (this is because the element has disappeared, and therefore can't trigger the `mouseleave` event). I fixed this bug by adding this option to hide the tooltip when the target element is clicked.


wwayne:
@lukeknep Does global method `ReactTooltip.hide()` work for you?

PR #99, opened 2016-05-11
Simplify setting Sstate in the showTooltip

Kiwka:


wwayne:
@Kiwka Good catch, thanks 👍

PR #102, opened 2016-05-13
Clear hide timeout on component unmount

### LINKED ISSUE #101: hideTooltip is fired when the component already unmounted
I have a component with lots of active cells with tooltips.Tooltips are initialized with 3 sec hide delay:
```jsx
        <ReactTooltip
          offset={{ top: 10 }}
          border={ true }
          delayShow={ 0 }
          delayHide={ 3000 }
          html={ true }
        />
```
When the user hovers the cells there is lots of hide timeouts fired and if the user navigates to another page (ReactTooltip is being unmounted) the timeouts are start firing and throws the errors into console:
ss 2016-05-13 at 1 34 56 pm
I see that it happens because of `setState` in timeout callback:
```js
  hideTooltip () {
    const {delayHide} = this.state
    clearTimeout(this.delayShowLoop)
    setTimeout(() => {
      this.setState({    // ← component is unmounted when the callback fired
        show: false
      })
      this.removeScrollListener()
    }, parseInt(delayHide, 10))
  }
```
So I gonna to introduce `delayHideLoop` variable to store the hide timeout id and clear it when the component is unmounting

oluckyman:
Resolves #101

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,6 +71,7 @@ class ReactTooltip extends Component {
       isCapture: props.isCapture || false
     }
     this.delayShowLoop = null
+    this.delayHideLoop = null
   }
 
   /* Bind this with method */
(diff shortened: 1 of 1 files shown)


wwayne:
@oluckyman 👍  thanks, merging!

PR #103, opened 2016-05-16
Allow user to specify separate off event for the tooltip

ondy1985:
If the user wants to hide the tooltip using a different event than the one used for showing it, he can use the **data-event-off** attribute.


src/index.js
```diff
@@ -118,14 +119,25 @@ class ReactTooltip extends Component {
...
         targetArray[i].setAttribute('currentItem', 'false')
       }
       dataEvent = this.state.event || targetArray[i].getAttribute('data-event')
       if (dataEvent) {
-        targetArray[i].removeEventListener(dataEvent, this.checkStatus)
-        targetArray[i].addEventListener(dataEvent, this.checkStatus, false)
+        dataEventOff = this.state.eventOff || targetArray[i].getAttribute('data-event-off');
+        // if off event is specified, we will show tip on data-event and hide it on data-event-off
+        if (dataEventOff) {
+          targetArray[i].removeEventListener(dataEvent, this.showTooltip);
```
wwayne: I don't think this can work because if you check the method `checkStatus`, you will find there are more things than showTooltip and hideTooltip. So it would be better if dataEventOff can work well with `checkStatus`(you don't have to use that method, I just mean maybe you can refer to it)

ondy1985: I haven't tested it thoroughly, but so far it works :-) To be honest, I don't fully understand the checkStatus method. Seems like it's there just to make sure that when the "event" happens, it closes the tooltip if it is opened, or opens it if it is closed, which you don't need when the opening/closing is bound to different events.

wwayne: Yeah, pretty much, but checkStatus also creates a listener for `when click other place, the tooltip should be removed`, so I don't know if dataEventOff can compatible with this feature or maybe we don't need that feature if user has set dataEventOff. I need to think about it and advices are welcome.

wwayne:
Could you resolve the conflicts, I plan to merge this PR and do some modification, thanks

ondy1985:
Hi, conflict is resolved.

ondy1985:
Hi, conflict is resolved.

wwayne:
Merging, thanks, will release new version soon

PR #104, opened 2016-05-18
Ensure data-html is converted to bool type

antoniodgonzalez:
When using data-html attribute, this value was assigned to the html state attribute without ensuring the bool type, so it was being assigned as a string "true" or "false".

Later, this value was accessed with the code:
`if (html) {`
this was evaluating always to true and causing the content to not being encoded.

Applied code to ensure bool value, similar to the code applied for the data-border attribute.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -267,7 +267,7 @@ class ReactTooltip extends Component {
       type: e.currentTarget.getAttribute('data-type') || this.props.type || 'dark',
       effect: e.currentTarget.getAttribute('data-effect') || this.props.effect || 'float',
       offset: e.currentTarget.getAttribute('data-offset') || this.props.offset || {},
-      html: e.currentTarget.getAttribute('data-html') || this.props.html || false,
+      html: e.currentTarget.getAttribute('data-html') === 'true' || this.props.html || false,
       delayShow: e.currentTarget.getAttribute('data-delay-show') || this.props.delayShow || 0,
       delayHide: e.currentTarget.getAttribute('data-delay-hide') || this.props.delayHide || 0,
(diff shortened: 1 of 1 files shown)


wwayne:
If the user wants to show a html tooltip on a specific element, for example:
```
<p data-tip data-html='<a>tooltip</a>'></p>
<ReactTooltip />
```
it won't show anything because `e.currentTarget.getAttribute('data-html') === 'true' || this.props.html || false` will return false

antoniodgonzalez:
From the documentation, I understood that this parameter should be boolean (true/false)
This same example appears in the documentation as:
```
<p data-tip="<p>HTML tooltip</p>" data-html={true}></p> or <ReactTooltip html={true} />
```

By reviewing the part of the code that handles this option, it seems that it is always using the placeholder, encoded or not depending on the html boolean value.

```
    if (html) {
      return (
        <div className={tooltipClass + ' ' + extraClass} data-id='tooltip' dangerouslySetInnerHTML={{__html: placeholder}}></div>
      )
    } else {
      const content = this.props.children ? this.props.children : placeholder
      return (
        <div className={tooltipClass + ' ' + extraClass} data-id='tooltip'>{content}</div>
      )
    }
```

wwayne:
Hmm. my bad, I didn't check it carefully, yeah you're right, thanks

antoniodgonzalez:
Thanks :)

Related to this, I noticed that other number attributes `data-delay-show` and `data-delay-hide` could potentially have the same issue, as they are not being converted to number. But didn't check myself and not sure if it is an actual problem.

wwayne:
It's ok, because they have been converted to Int before using

Issue #16, opened 2015-06-17
Translate3d on the parent (not necessarily direct parent) causes the tooltips to misalign and fire incorrectly in on Safari 8.0.6 (latest)


bluejamesbond:

Moussawi7:
@bluejamesbond ,
Same issue on Desktop browsers(FF,Chrome...).
Did you solve this issue?

adamscybot:
This is a bug in the spec of transforms where a new coordinates system is created for every element that is transformed.

In theory, you could work around by calculating the position of the parent and passing the values to the offset prop.

CremAlex:
It seems when you refactored your code you removed the fix provided by the pull request. I upgraded from 3.0.0 to the latest and a simple translate3d(0,0,0) make the tooltip position bug again.

![image](https://cloud.githubusercontent.com/assets/5360522/16808150/9521a7b4-491b-11e6-82ba-647394e39ee9.png)

wwayne:
@CremAlex Good to know `npm unpublish`, it would be super useful for me, thank you. As for the translate3d fix, I was planning to improve that part, but I missed it, so please use old version for the moment, I will let you know when I fix this bug.

wwayne:
@CremAlex This has been released with `3.0.11`

CremAlex:
Thanks ! Just a question why not check if the current element is under a transform aswell? 

In my case I have this : 

    <div className={ Button } style={ { transform : /* some react-motion transform */ } } data-for={id} data-tip>
        <i className='icon'/>
    </div>

And I don't feel like adding an extra div just for my tooltip when we can check the current node element before checking the parent(s).

wwayne:
Released `3.0.12`, looks npm can't republish with same version

CremAlex:
Oh well I checked the doc and yeah it looks like you can't. With sinopia I can do it at work

kpavankotesh:
I see this issue in IE11, any solutions?

PR #106, opened 2016-05-25
Workaround for fixed elements inside transformed elements

nd0ut:
Ref #16


wwayne:
Could you introduce background about your thought? I see there is an new attribute `checkParentTransform`, and what you use this attribute to do?

nd0ut:
@wwayne Yeah, sorry. The main idea is that the tooltip will scan it's own parent chain for the elements being transformed with css transform. I there is such element then the tooltip will be positioned relative to that element.

And I think it should be enabled by default. And I forgot why I did that new prop :smile:

danbrianwhite:
@wwayne Any updates on merging this pull request? The update by @nd0ut fixes an issue I have with tooltip position in a fixed element.

wwayne:
Sorry I've forgotten this... could you please resolve the conflict so that I can merge this in the future?

nd0ut:
Hmm it seems I'v did something wrong. There are some strange changes after building.

wwayne:
@nd0ut Never mind, I will fix it. But I need to check the code first... anyway, thanks for your PR

wwayne:
@nd0ut Thanks, will do some modification before next release 👍

PR #129, opened 2016-07-03
Feature: Support custom attribute for tooltip content

gottsohn:
#### What does this PR do?
 - It adds an extra prop to the component to receive a custom handle, used to target the tooltip content i.e. custom props can be used in place of `data-tip`.
 - The prop is called `handle` and assumes value `'data-tip'` as the default value when the prop is not used.


wwayne:
I don't think this is necessary because too many choices may lead confusion( if he happens to use `data-html` to replace `data-tip` which should be invalid ), and looks people are get used to using `data-tip`. Let's closing this PR unless lots of people ask for this.

PR #131, opened 2016-07-07
Add possibility to recalculate content on show tooltip

pokidovea:
Problem:
A tooltip should show content, wich changes over time. E.g. show time until any event.

Solution:
Pass a function to tooltip,  which will be called on showing the component. This function will return needed content, that will be rendered instead of children.


wwayne:
Thanks, I will do some small fix before next release

PR #139, opened 2016-07-16
Fix jittering when on move while effect is "solid".

mattzeunert:
Fix for https://github.com/wwayne/react-tooltip/issues/138.

I also noticed that removing the listeners wasn't working correctly. `::this.showTooltip` creates a new function that's different from `this.showTooltip`, so it can't be removed with `.removeEventListener('mouseenter', this.showTooltip)`.


wwayne:
I think the function has been correctly removed, the most easy measure to test it is put a `console.log` into the `showTooltip` function to see if it would output only once every time showing the tooltip.

This problem only happens in HD screen, retina for example, someone mentioned this before, I didn't think it's obvious under effect `float` so I closed the ticket, but for solid, it would be a problem, will check this later.

mattzeunert:
Thanks. I checked with `getEventListeners` in Chrome and it showed lots of them.

Sorry about the build commit by the way. I started to load the forked branch locally with NPM and didn't realize doing a build on it would update the PR.

wwayne:
Some PR to this component includes built files, so it's acceptable actually : )

PR #140, opened 2016-07-16
Make use everything as a tooltip example clearer.

mattzeunert:
It wasn't immediately clear to me what this allows you to do. Including some HTML/styling in the tooltip would make it easier to understand.


mattzeunert:
Actually, I don't think this makes much sense, since the tooltip already allows HTML without using `data-for`.

PR #167, opened 2016-08-02
add type dark to the readme

sjsnider:
-I wanted a dark tooltip, but didn't find what I wanted after trying all the 'type' options listed in the readme, so started going through index.scss to see how I was going to manipulate the css and there was what I was searching for all along!  type: dark   Figured I'd just add it to the readme real quick


wwayne:
Actually if you don't add `type` attribute, the tooltip will use `dark` as its default type, and I think it's easy to find these out by checking the example http://wwayne.com/react-tooltip/, so I don't think it is necessary to add this to the README.

sjsnider:
ah, naturally I never tried that.  Thanks for letting me know

PR #170, opened 2016-08-03
Issue 157 - adding callbacks when hover on and off the tooltip.

13DeAnda:
Adding functionality to allow a user to create callbacks for when they hover on and off an item to show the tooltip.


wwayne:
I think the callback function should be added into `setState({}, () => { cb() })`, or the callback function will be called before real `show` and `hide`, especially when user want something like delayShow or delayHide

Issue #144, opened 2016-07-21
Keep tooltip open when mouse hovers over


rollsroyc3:
I am using HTML with a mailto link in the tooltip that I want the user to be able to click on. Is is possible for the tooltip to stay open while cursor is hovering? I saw that there is a hide delay but I think that would just confuse the users.

Thanks for the great plugin!

wwayne:
At the moment, I would suggest you use the delayHide(you can see the example page for the usage), I may add new attribute in the future to support this

rollsroyc3:
Ill probably just leave it as plain text for now then. If a user moves cursor away and tries to do something else but the tooltip is still up for 3 seconds blocking other inputs its not very user friendly. If I ever get the time maybe I will fork and try to add it.

huumanoid:
I think it's important to mention for someone who wants hoverable tooltips and looking for solution.

There is great workaround published by @wwayne himself! 
http://wwayne.com/react-tooltip/ check out "Theme and delay"

As for me, it has taken some time before i realized that there IS a solution, so such comment would have been useful for me.
Thanks @wwayne for the great plugin!

akaSybe:
@huumanoid could you please share workaround that was published on the website that you mentioned in comment above? This website is unavailable now and I need hoverable tooltips =\

akaSybe:
I've found it, it is on github pages: http://wwayne.github.io/react-tooltip/

Rogger794:
It supports already data-tip='custom show' options. I would close this issue.

dipeshparashar:
@roggervalf How to keep it open. Can you elaborate. I have textarea inside the tooltop, i want to keep the state textarea value and the tooltip keeps closing during the state update

dewaleolaoye:
@roggervalf on this issue `data-tip` which is now `data-tooltip-content` in v5 only accepts `string`, can we have it to accepts `jsx` as well? or is there a work around it

harrybasra95:
@dewaleolaoye Just found out that there is `clickable` prop which can be passed to achive this.

MahmoudHafez1:
Add this class to tooltip and let it clickable

.custom-tooltip {
  pointer-events: auto;
  &:hover {
    visibility: visible !important;
    opacity: 1 !important;
  }
}
 <Tooltip clickable={true} className="custom-tooltip">

PR #171, opened 2016-08-03
Feature issue144

13DeAnda:


wwayne:
I think #144 is asking for hover on tootlip which can be implemented but somewat complex for the moment, you can check the example in `Theme and delay`

As for click trigger tooltip, it has already been implemented by custom event

PR #179, opened 2016-08-16
Add bower.json

### LINKED ISSUE #97: need bower support
I was trying to install this package with bower. Couldn't install the package. Always got  

ECONFLICT Unable to find suitable version for react-tooltip

mikkopiu:
Fixes #97 

- Created `bower.json` in project root.
- `main` is set to `standalone/react-tooltip.js` to allow tools like [wiredep](https://github.com/taptapship/wiredep) to automatically locate the correct distribution file
- React and ReactDOM moved to `dependencies` as Bower does not have `peerDependencies`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,33 @@
+{
+  "name": "react-tooltip",
+  "description": "react tooltip component",
+  "main": "standalone/react-tooltip.js",
+  "moduleType": [
+    "globals",
+    "amd",
+    "node"
+  ],
+  "license": "MIT",
+  "ignore": [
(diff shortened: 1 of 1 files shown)


wwayne:
@mikkelvagn Cool, really thanks, I will merge this asap. I suppose that I should also publish this to bower, right?

mikkopiu:
Yeah, that would be great -- I'm not actually sure if it is necessary, but it sure wouldn't do any harm.

wwayne:
Thanks, looks it is already on the Bower

PR #183, opened 2016-08-19
Remove react-dom from Bower dependencies

mikkopiu:
Hey, I noticed a mistake in my earlier PR about adding a bower.json (https://github.com/wwayne/react-tooltip/pull/179): `react-dom` is not a proper Bower dependency and the `react` package instead comes with `react-dom` properly configured in its `main`: https://github.com/reactjs/react-bower/blob/master/bower.json#L3

Sorry about the confusion, but here a fix PR.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,6 @@
     "url": "https://github.com/wwayne/react-tooltip"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


wwayne:
Cool thanks 💯

Issue #186, opened 2016-08-31
Use tooltip for form validation


oyeanuj:
Hi @wwayne, back with some more questions :)

I am trying to use the tooltips to show validation error messages for my forms, like in the screenshot below. Essentially, once a user clicks on the 'Submit' button, if there is an error, then show it in the tooltip, in context.

<img width="429" alt="screen shot 2016-08-31 at 4 39 08 pm" src="https://cloud.githubusercontent.com/assets/9633371/18150160/9c0c4bfc-6f99-11e6-96bb-6d2cb38f9e40.png">

I was thinking to use the static methods to show it in case of errors, but I am running into a couple of issues while trying to implement this.

- How to invoke the method to invoke the tooltip conditionally? Usually, in forms, the `onClick` is already being used for submitting the form? I would like it be shown after `onClick` if there is an error? In the code below, I try to do the same but it doesn't show up the first time I click on the button. I have to click on the button again for the tooltip to show up.

```javascript
//Outside the render function
  showTooltip = (e) => {
    console.log(e.target);
    ReactTooltip.show(e.target);
  };

//First solution attempted: In the render function
<ReactTooltip getContent = {showError} effect = "solid" />
<Button
  ref = "flagButton"
  data-tip
  data-event = "focus" //have also tried "data-event = "click focus"

  disabled = {submitting}
  label = {submitting ? "Sending" : "Send Report" }
  submitting  = {submitting}
  type = "submit"
  onClick = {error && anyTouched && this.showTooltip}
/>
```

- I tried making `data-tip` conditional which almost worked but it shows an empty tooltip when there should be no tooltip. Code below

```javascript
//Second solution attempted: In the render function
<ReactTooltip getContent = {showError} effect = "solid" />
<Button
  ref = "flagButton"
  data-tip = {error && anyTouched}

  disabled = {submitting}
  label = {submitting ? "Sending" : "Send Report" }
  submitting  = {submitting}
/>
```

Is there any other alternative that you can think of to invoke the tooltip in such a case? Essentially, is there a way to make the tooltip **render conditionally**?

- Secondly, is there a way to make the tooltip disappear after a few seconds?

Any thoughts would be very appreciated, thank you!

oyeanuj:
Updated the comment to remove issue that was my bad.

wwayne:
@oyeanuj I add a new option `disable` which would useful in your situation, so for your code, `ReactTooltip.show` should invoked after submit and fail, for example:
```
[...]

PR #189, opened 2016-09-09
accept null from getContent

cklab:
In response to issue #186 

Return `null` from `getContent` to hide tooltip.


wwayne:
Good catch, but `typeof null === 'object'`, so I've sent another PR that can support both undefined and null. https://github.com/wwayne/react-tooltip/pull/202

Anyway, thanks for your suggestion and PR 👍

cklab:
@wwayne That's correct, and that fact is exactly why this PR works. I'm confused as to your reasoning, this and #202 are accomplishing the same exact thing...

wwayne:
For the same goal but not the exact thing, I dealt with getContent and children separately and worked with a new state which introduced from previous PR to control it so that it can cover more edge cases.

No offend, but for this PR

1.  it can't deal with `getContent={() => { return }}`
2.  it can't deal with `<ReactTooltip>{}</ReactTooltip>`
3.  it doesn't work when working with `delayHide`, for example 
```
<a data-for='custom-class' data-tip='hover on me will keep the tootlip'>(･ω´･ )</a>
<a data-for='custom-class' data-tip getContent={() => { return null }}>(･ω´･ )</a>
<ReactTooltip id='custom-class' class='extraClass' delayHide={1000} effect='solid'/>
```
4. I think we don't need to add `typeof getContent !== 'undefined'`, I suppose `if (getContent)` is enough here

PR #190, opened 2016-09-12
Fixes IE bug when empty data-tip.

arnulfholm:
How to see the bug:

1. Create two elements with data-tip attribute
2. Add content to one of the data-tip attributes, and have en empty string in the other
3. Hover over the one with content
4. Exit the element
5. Hover over the one without content

Result:
An empty tooltip will appear in IE 10 and 11 (probably others as well)
If exit the element without content, then enter again, all is ok.

Probable reason:
Since the content of the tooltip is not reset after exit, some logic probably thinks there is content, even though the new item does not have content.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -312,7 +312,8 @@ class ReactTooltip extends Component {
     const resetState = () => {
       const isVisible = this.state.show
       this.setState({
-        show: false
+        show: false,
+        placeholder: ''
       }, () => {
         this.removeScrollListener()
         if (isVisible && afterHide) afterHide()


wwayne:
Thanks for exploring the problem, but I found it's little more complicated than what we thought. For example, the problem would be still there if we are using `delayHide`. 

Finally I add a new state `isEmptyTip` to control this situation and sends a PR https://github.com/wwayne/react-tooltip/pull/201.

Anyway, thanks for your findings and PR.

PR #195, opened 2016-09-20
Use node parent when calculating offset

iamdoron:
Hi,

I added some transforms to the examples as a quick way to see it's working. When calculating the tooltip position, the parent of the node should be taken as an offset, and not the target. 

This is not a complete solution, because the node can be within multiple chain-of-parents with transforms (which have to be accumulated), but at least now if the target is inside a parent with transform the tooltip is in the right position.

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -45,7 +45,7 @@ const Test = React.createClass({
       <div>
         <section className='tooltip-example'>
           <h4 className='title'>React Tooltip</h4>
-          <div className='demonstration'>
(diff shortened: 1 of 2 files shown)


wwayne:
I think we can't use node to replace the target here, because the logic doesn't make sense. And I can find another test case that makes this out of work... so I'm still thinking about it and trying to figure out a more clear logic.

Anyway, thanks for your PR and exploration, I'm closing this.

wwayne:
After spending almost one hour figuring out the logic behind this, the only thing I found is you are right, thanks 👍

PR #198, opened 2016-09-28
Freeze the tooltip through a property

wbazant:
This pull request introduces an ability to "freeze" the tooltip through external prop change, which is meant to happen through the React component lifecycle. When the tooltip is frozen, the updates (show, move, disappear) do not happen and instead get stored, and the last of these updates happens when the tooltip is unfrozen. 

[Demo](https://wbazant.github.io/react-tooltip/)

The feature lets you put interactive features in your tooltip, like buttons. 

In Expression Atlas we wanted to use the UX pattern of hovering over an element displaying the tooltip, and clicking on the element stabilising it so that you can interact with it.


wwayne:
I personally don't want to add this feature to the library, because:

1. It is somewhat custom and it needs to think about more custom situations. For example, I'm using a chrome plugin called Vimium which will bind keys for user to operate the website, so I'm not able to use `f` to freeze the tooltip until I stop the plugin.
2. User can use delayHide to achieve same purpose(I'm trying to optimize it so that others can easy understand and use it).
 
Thanks for your PR and idea, it looks good for your team, but maybe not suit to be an option of the library. I'm closing the PR.

PR #207, opened 2016-10-12
Return tooltip to original position when possible

hassanbot:


wwayne:
I'm not sure we should add this option, in which situation you need such function?

wwayne:
Okay I just checked Youtube slider, and I suppose that you are adding an option to keep the tooltip alway in a same position? no matter if it is out of the screen or not?

wwayne:
Hm, thanks for the detailed explanation, I'm clear now.

But I don't think it is a good idea to add an new option for reverting the tooltip to its origin, I suppose it would be better if the tooltip can revert to its origin automatically instead of keeping new position? I'm not sure how to implement this, but I will take investigation later.

huumanoid:
Hey guys I'm really impressed by this fix, I think this fix is very important.

Here is another approach to achieve desired behavior, please, take a look!
https://github.com/huumanoid/react-tooltip/blob/fix-original-position/src/index.js#L385

huumanoid:
After a while, I can't said why my solution is better that @hassanbot's one. @wwayne, please clarify what do you want from a good solution?

it seems that `getPosition` can't automatically switch to the origin because it needs information about the origin. So, anyway we have to handle somehow that tooltip has the original `place`. @hassanbot handles it in the `getPosition`, and I'm handle it in the `updatePosition`.

aronhelser:
@hassanbot @huumanoid do you have a preference here? I've tested hassanbot's solution, and it works, and it appears a bit simpler than huumanoid's solution. I'm inclined to merge this PR.

wwayne:
@aronhelser Please go ahead if it is good for you, thanks 👍

smakosh:
It doesn't actually work as the tooltip changes it's position from bottom to left

Issue #210, opened 2016-10-19
Tooltip inside a <p></p> results in a warning


dropfen:
It actually works, but the default tooltip wrapper is a `<div>`
And this makes it problematic to use it within a `<p>` tag.. In our case we need a tooltip for an img which is inner p tag.. we use span based template to show the tooltip content, but we get a warning message because the tooltip wrapper is a div.

`Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>. See ReactMarkdown > p > ... > Image > ItemIconWithTooltip > ReactTooltip > div.`

wwayne:
Thanks, merged the PR, will release in 3.2.4

joaoreynolds:
For future readers - to save the few minutes it took me to go read the docs. You can pass in the option `wrapper="span"` (default remains `div`) which will get around this error.

Aryaman73:
Not sure if this is working as intended, but it looks like with `wrapper="span"` there is a p tag there, and I'm getting a similar error: 

![image](https://user-images.githubusercontent.com/53343807/127930973-cd322b00-0b34-424e-ad9f-0e879de539ca.png)

I'm using `v4.2.21`

MilanLund:
I am experiencing this issue too `v5.7.1`. 
Using with Next.js App Directory inside a client component.

PR #211, opened 2016-10-19
Dynamic wrapper

dropfen:
Adds a new prop _wrapper_.
default: "div"
supportedWrappers: span, div

Let the user define the wrapper for the Tooltip dynamically.
fixes -> #210


wwayne:
I was thinking why don't just use `<span>` directly, would it be better?

dropfen:
Well, yeah you can use *span* as the default, but it could mean some upgrading issues, for people who are rendering `divs` in production. It's safer to keep div as the default.

MatanBobi:
After reviewing this PR and its result,
I keep getting <wrapper> elements inside my html.
I think that its because of this:

[https://github.com/dropfen/react-tooltip/blob/25aa42a2b8113aff21bb053f309d7ace9b937f93/src/index.js#L428](url)

React doesnt render the variable wrapper as a div/span but as a DOM element.
I think its because that the variable isnt with capital letter so react doesnt see it as a component.

huumanoid:
Yes, you are right! Thank you for your review!
Babel react preset treats `wrapper` as a tag name, not as a component.
https://github.com/wwayne/react-tooltip/blob/master/standalone/react-tooltip.js#L843

If you change `wrapper` to `Wrapper`, everything is fine.
Could you make a PR, please?

Also, take a note that there is a linter error.
```
~/github/react-tooltip/src/index.js
  441:9  error  Empty components are self-closing  react/self-closing-comp
```

PR #216, opened 2016-10-27
Fix fire of global show event for IE

SubVersive:
For IE create new Event object and set details separately because initEvent doesn't support this and we expect lately to have detail property on event object.


wwayne:
Thanks 👍

PR #220, opened 2016-10-29
Use parameter 'useCapture' for remove listener if it is registered as…

achernetsky:
Use parameter 'useCapture' for remove listener if it is registered as a capturing


achernetsky:
Hello!

What do you think about this pull request?

Thanks.

wwayne:
I think this is good, thanks : )

PR #222, opened 2016-11-04
Fix issues with tooltip in scrollable div and IE11 fixed position - Added feature to pass styles

danbrianwhite:
Fix issues with tooltip in scrollable div and IE11 fixed position.
Added feature to pass styles (important to override some weird bugs in IE11 when using transforms on parent elements)
Updates a few spellings

Git history is a little ugly, but left it to show updates by contributors.


danbrianwhite:
@wwayne can you review this pull request and let me know you are planning on merging it? 

I would like to use your master branch of the project instead of a fork.

wwayne:
Sorry I was busy with other things, so what's the problem you met with in scrollable div?

src/index.js
```diff
@@ -44,7 +45,8 @@ class ReactTooltip extends Component {
     afterHide: PropTypes.func,
     disable: PropTypes.bool,
     scrollHide: PropTypes.bool,
-    resizeHide: PropTypes.bool
+    resizeHide: PropTypes.bool,
+    style: PropTypes.object
```
wwayne: what's the style props used for?

danbrianwhite: In IE11, if the parent element for the component for the tooltip has a transform,translate(-50%, -50%), then the tooltip will not render in the correct location. By enabling a style prop to be passed down to the tooltip, a developer then can resolve the issue by passing a style prop with {position: 'absolute'} for the tooltip.

danbrianwhite:
The problem I ran into with the scrollable div was that when you scroll in a div, it does not pass down the event to the window. When this happens, the tooltip does not disappear in the scrolled div. 

By using the scrollparent package, the listener is attached to the closest scrollable parent element and when scrolling, the tooltip will correctly disappear.

aronhelser:
Do you still need this PR merged?

PR #223, opened 2016-11-04
Fix/security csp style src

briantrice:
My team uses CSP to secure a React app that uses ReactTooltip, but the way the style header is injected is incompatible, triggering a security warning about dynamic javascript injection.

This adds another build step that generates a separate static CSS file for inclusion, with a global configuration flag to avoid the injection.

No defaults have been changed, so the current example and usage story work the same, but I'm hoping a reasonable way to toggle the flag can be made other than to pass a prop `insecure={false}` to every component instance.

What do you think?


paulyoung:
@wwayne is there anything I can do to move this one along?

wwayne:
This is awesome, thanks

sihil:
This is a great addition (just hit the CSP issue) @wwayne. Having said that I'm not a big user of the client side ecosystem and can't find the docs on how to explicitly include the CSS. Have copy and pasted for now, but suspect there is a better way. Any chance someone could improve the docs with a few lines on how to set `insecure=false`?

PR #229, opened 2016-11-23
Replaced Object.keys(...) by Object.getOwnPropertyNames(...)

gauthierj:
Hi,

We replaced Object.keys(...) by Object.getOwnPropertyNames(...) because it's safer and has same browser support. 

It's safer because it will only retrieves index of the NodeList, and not methods. it is no longer needed to filter the output.

We did this because we encountered a problem in regard with this.

Best regards,

Gauthier JACQUES


wwayne:
@gauthierj Thanks for your modification and explanation : )

PR #234, opened 2016-12-05
Fix/deploy standalone css

briantrice:
Includes the static file CSS in the deploy and standalone builds.


codinronan:
@wwayne if you wouldn't mind including this?

I'm dumping the SASS file from source right now into my own project but this would be a nice bump :)

huumanoid:
Hi guys! Ok, I see that there is `standalone/style.css` and `dist/style.css` files now.
Please, explain me, why do you need these file to be in the distribution?
`react-tooltip` styles are injected to the `/html/head/style[@id="react-tooltip"]`. I guess I'm missing some use cases, please, tell me!

@codinronan don't you need the `.scss` file to be in the distribution too?

darekfrynas:
Hey @huumanoid, with standalone CSS file it would be much more convenient to overwrite themes, without being forced to use `property: value !important` notation.

briantrice:
I am no longer currently in need of the `react-tooltip` library, so won't advocate for this.

aronhelser:
Closing until someone requests it.

PR #236, opened 2016-12-08
escape backslash in getTargetArray method

rnons:
it's a rare case, but if `data-tip` contains backslash, `getTargetArray` or more specifically `querySelectorAll` won't find the element

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -150,7 +150,8 @@ class ReactTooltip extends Component {
     if (!id) {
       targetArray = document.querySelectorAll('[data-tip]:not([data-for])')
     } else {
-      targetArray = document.querySelectorAll(`[data-tip][data-for="${id}"]`)
+      const escaped = id.replace(/\\/g, '\\\\').replace(/"/g, '\\"')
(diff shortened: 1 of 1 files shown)


rnons:
@wwayne can you take a look a this? Thanks

huumanoid:
Hey @rnons sorry for long response.
Thank you for this interesting case and patch for it! I think your PR looks fine.

How about to make this PR more general?

According to [https://www.w3.org/TR/html5/dom.html#the-id-attribute](HTML5 specification), the `id` could contain any character except space characters.

I see another case: when "id" contains double quotation mark, `document.querySelectorAll("[data-for="hello"world!"]"` won't find anything.
We should escape `"` too: `document.querySelectorAll("[data-for="hello\\"world!"]` (success!)

Any other cases? Please, tell me, if you find one!

rnons:
@huumanoid updated

wwayne:
@huumanoid Sure, will do today, thanks @rnons

wwayne:
Merged and released as 3.2.9, it should be 3.2.8, but some error happened when I did `npm publish` and I was not sure if it was published successfully... So I release a new version for safe

PR #239, opened 2016-12-16
Allow to use custom callback on globalEventOff events

frsv:
Currently there is no way to fire globalEventOff events conditionally – for example, to hide tooltip only if Escape key is pressed. In this PR I added globalEventOffCallback function property, which first argument is original hideTooltip function, second argument is original event, so it become possible to hide tooltip only when it's really necessary:

```
const customCallback = function(hideTooltip, event, hasTarget) {
    if (event.keyCode === 27) hideTooltip(event, hasTarget)
}
...
<ReactTooltip globalEventOff="keyup" globalEventOffCallback={customCallback} />
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Global|Specific	|Type	|Values  |  Description
  event |   data-event  |  String  |  e.g. click | custom event to trigger tooltip
  eventOff |   data-event-off  |  String  |  e.g. click | custom event to hide tooltip (only makes effect after setting event attribute)
  globalEventOff | | String| e.g. click| global event to hide tooltip (global only)
+ globalEventOffCallback | | Func| (hideTooltip, event) => hideTooltip(event)| global event callback to hide tooltip. Works only if globalEventOff specified
  isCapture | data-iscapture | Bool | true, false | when set to true, custom event's propagation mode will be capture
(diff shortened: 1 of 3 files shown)


CohenBenco:
Exactly what i was looking for.. is it going to be merged soon?
Thanks!

frsv:
@wwayne do you have any thoughts about this feature?

huumanoid:
Hello, @frsv, I love this feature! 
Thanks to your PR, I've realized, that I need something the same, but a bit different - I need to conditionally **show** the tooltip! I think this can be implemented in the same manner as in the current PR, so, thank you, your idea is really great.

However, your code doesn't looks safe.
Are you sure that event listeners are properly removed?
Every time you invoke the `bind` method, the new function created.

Play it yourself!
https://jsfiddle.net/hmnid/7Lkrsj7g/ - wrong `bind` usage.
https://jsfiddle.net/hmnid/7377uen6/ - correct `bind` usage.

CohenBenco:
Hey guys, ive also just noticed that if the tooltip is open by click and
stay open, its scrolled with page because its being positioned as fixed..

בתאריך 1 במרץ 2017 19:49,‏ "huumanoid" <notifications@github.com> כתב:

huumanoid:
@CohenBenco I suppose that the problem indeed exists, but I'm not sure this bug is related to this PR.
There are some issues with hiding on `scroll` event.

For example, I'm experiencing such an issue in case I've set `delayHide` to `1000`. Do you have the same case?

aronhelser:
Sorry we couldn't get this one in....

Issue #246, opened 2017-01-29
Tooltip with delayHide disappears after re-hovering


huumanoid:
Let's say we have
* a `ReactTooltip` component with `delayHide={1000}`
* one or several elements associated with this tooltip

When we hovering one of this elements and then hovering another (or re-hovering the same) element before first tooltip has gone, the new tooltip disappears as soon as the `delayHide` timer, associated with first the tooltip, has elapsed.

![reproduction of the described issue](https://cloud.githubusercontent.com/assets/3264579/22401908/678e379a-e5f6-11e6-901c-6cceca5cc1f8.gif)

It may seems that the problem is insignificant, because even if tooltip has gone, once you have moved mouse pointer, the tooltip appears again (as well as at my first gif).
Unfortunately, it's not.

When you have rebuilded your tooltips using `ReactTooltip.rebuild()`, the behavior changes to undesirable.
![same as first, but after ReactTooltip.rebuild()](https://cloud.githubusercontent.com/assets/3264579/22401941/e9728b16-e5f7-11e6-8d1d-917751f7753c.gif)

I have faced this issue while implementing a table with hoverable tooltips attached to table's cells.
Hoverable tooltips implemented according to ["Theme and delay" section of examples page](http://wwayne.com/react-tooltip/)
When user moves mouse pointer from up to down over the table's cells and holds pointer over cell they interested in, tooltip disappears and this is discouraging.

You may reproduce first part of this issue even at ["Theme and delay" section of examples page](http://wwayne.com/react-tooltip/).

My test stand with handy "Rebuild ReactTooltip" button is available at http://hmnid.ru:8090/
and also available on github https://github.com/huumanoid/react-tooltip-delayhide-issue-test

PR #247, opened 2017-01-29
Fix delayHide issue #246

huumanoid:
# Description 
The following text assumes that reader has read description of issue #246.

At first, I've just added `this.cleanTimer()` in `showTooltip` method.
Problem described in #246 has completely gone. 

But it's reasonable to ask: why behavior is different after `ReactTooltip.rebuild()`?
It was caused by invalid effect detection in `bindListeners` method. 

I think, this should be considered as a bug:
effect

In the example above, there are one `<ReactTooltip />` component with default setup, both buttons SampleText(1|2) attached to this `ReactTooltip`, `data-effect` of the second button has been set to `solid`.

I haven't get an idea why `isCapture` is implemented as decorator, but I've followed the same way and have added `getEffect` decorator. I wonder what you think about it.

# Tests
Test stand still available at http://hmnid.ru:8090/ . Test stand for `effect` issue available at http://hmnid.ru:8090/effect.html

Demo with fixed react-tooltip available at http://hmnid:8091/ and http://hmnid.ru:8091/effect.html respectively.

Sources of demos available at https://github.com/huumanoid/react-tooltip-delayhide-issue-test

### DIFF
--- a/src/decorators/getEffect.js
+++ b/src/decorators/getEffect.js
@@ -1,0 +1,11 @@
+/**
+ * Util method to get effect
+ */
+
+export default function (target) {
+  target.prototype.getEffect = function (currentTarget) {
+    const dataEffect = currentTarget.getAttribute('data-effect')
+    return dataEffect || this.props.effect || 'float'
+  }
+}
+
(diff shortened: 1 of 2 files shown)


wwayne:
Really thanks for your work and detailed explanation :), I'm gonna build and release new version, it should be `3.2.3`

huumanoid:
Very glad to see that you have already reviewed it and pushed it to npm. Thank you! It's important for my project.

PR #254, opened 2017-02-07
Add support for className prop

jonbeller:
Solution for https://github.com/wwayne/react-tooltip/issues/242

I chose `className` as the name of the prop to match the convention of how CSS classes are specified on React components. It is also necessary to accept `className` for compatibility with the library [styled-components](https://github.com/styled-components/).

I did not remove the ability to call the prop `class`, but I updated the documentation and examples to prefer `className`. Let me know if you want me to add back references to `class`!


wwayne:
Good to me, thanks :), will deploy to npm late of today

Issue #256, opened 2017-02-09
undefined is not a function (evaluating 'e.getAttribute("data-iscapture")')


LeoIannacone:
This code react-tooltip/dist/decorators/isCapture causes error in Safari (v8 and v9 for sure):
```
exports.default = function (target) {
  target.prototype.isCapture = function (currentTarget) {
    var dataIsCapture = currentTarget.getAttribute('data-iscapture');
    return dataIsCapture && dataIsCapture === 'true' || this.props.isCapture || false;
  };
};
```

wwayne:
I just tested on Safari  `9.1.3 (11601.7.8)`, didn't find the error, any clues?

LeoIannacone:
Not really,

I get this error with safari v9 (iOS) and v8 (Mac 10.10)

Everything is fine with version 3.1.7

huumanoid:
Could you provide more detailed stack trace, please?
Is problem exactly browser specific? Have you tried another one?

LeoIannacone:
Stack trace:
```
TypeError: undefined is not a function (evaluating 'e.getAttribute("data-iscapture")')
  at isCapture(~/react-tooltip/dist/decorators/isCapture.js:9:0)
  at isCaptureMode(~/react-tooltip/dist/index.js:214:0)
  at forEach([native code])
  at isCaptureMode(~/react-tooltip/dist/index.js:214:0)
  at isCaptureMode(~/react-tooltip/dist/index.js:214:0)
  at setValueForProperty(~/react-dom/lib/DOMPropertyOperations.js:119:0)
  at exports(~/react-dom/lib/ReactVersion.js:13:0)
  at closeAll(~/react-dom/lib/escapeTextContentForBrowser.js:36:0)
  at perform(~/react-dom/lib/escapeTextContentForBrowser.js:36:0)
  at perform(~/react-dom/lib/escapeTextContentForBrowser.js:36:0)
  at exports(vendor.js:11662:23)
  at exports(vendor.js:11662:23)
  at ? ([native code])
  at closeAll(~/react-dom/lib/escapeTextContentForBrowser.js:36:0)
  at perform(~/react-dom/lib/escapeTextContentForBrowser.js:36:0)
  at batchedUpdates(~/react-dom/lib/ReactMultiChild.js:23:0)
  at exports(vendor.js:11662:23)
  at internalInstance(~/react-dom/lib/ReactUpdateQueue.js:24:0)
[...]

PR #260, opened 2017-02-16
Fix NodeList to Array convertion

huumanoid:
Related to issue #256 

Hope this PR would fix it once and for all.

Thanks @LeoIannacone for his diagnostics and issue report.
Thanks @tasiek for his extremely helpful diagnostics and feedback about my patch.
Thanks to @gauthierj for his important feedback, I'm glad that this approach OK for him.

Thanks everyone who has faced this issue and test this fix to make sure problem it gone.
```js
npm install 'huumanoid/react-tooltip#fix-undefined-is-not-a-function'
```

Discuss:
Should this library directly depend on `fbjs`? `react-dom` is already depends on `fbjs`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -155,6 +155,7 @@ class ReactTooltip extends Component {
       targetArray = document.querySelectorAll(`[data-tip][data-for="${id}"]`)
     }
 
+    // targetArray is a NodeList, convert it to a real array
     return createArrayFromMixed(targetArray)
   }
 
--- a/src/index.js
+++ b/src/index.js
@@ -16,6 +16,8 @@ import getPosition from './utils/getPosition'
 import getTipContent from './utils/getTipContent'
 import { parseAria } from './utils/aria'
 
+import createArrayFromMixed from 'fbjs/lib/createArrayFromMixed'
+
 /* CSS */
 import cssStyle from './style'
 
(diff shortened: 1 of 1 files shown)


wwayne:
@huumanoid Thanks a lot :), but the CI failed for `Error: Cannot find module 'fbjs/lib/createArrayFromMixed'`, do we need to add fbjs as dependencies then?

huumanoid:
Thanks for your response!
The situation is quite hard.
As mentioned at https://github.com/facebook/fbjs, it's not reliably to depend on fbjs. They don't follow semver strictly.
Both legacy and modern react are depend on fbjs.

I see three possible solutions.
1) Just add fbjs to dependencies. Con: it may break at any time.
2) Add strict dependency on the exact current version. Pro: breakage is less likely, however, is still possible, I guess. Con: when they update `fbjs`, we will have at least two copies of fbjs in `node_modules` (one for `react-tooltip` and one for `react-dom`. Who cares?
3) Just copy&paste required code to this project. Pros: reliably and lightweight. Cons: Copypasting of the already available code? Err... Moreover, are licenses compatible?

I can't do such a decision by myself.
Copy&paste is easy to do, required code has zero dependencies.
I trust in your wisdom. What should be do in this situation?

wwayne:
@huumanoid I prefer to check their solution first and write it by ourselves (well it's like copy paste, haha). I can do it in these 2 days.

Thanks again,  your [comment](https://github.com/wwayne/react-tooltip/issues/256#issuecomment-279196746) is really helped.

PR #272, opened 2017-03-20
Fix wrapper tag

huumanoid:
This PR related to #211

After applying #211, the wrapper DOM element for `ReactTooltip` expected to be `<div>` or `<span>`, but actually it's a `<wrapper>`:
screenshot from 2017-03-20 16-02-37

The reason of this behavior described by @MatanBobi [here](https://github.com/wwayne/react-tooltip/pull/211#issuecomment-287066189), and some more details are in c37e0635338b91c117f14273b4ac5d8b9ea5dd3a commit message.

Another issue is that we can't pass `React.DOM.div` to `React.createElement`, we should pass `'div'` instead. Read more: 3c85424c853f51b8295f081814f3eb59c2807695

Thanks @dropfen for #211 and for his concern about backward compatibility.
Thanks @MatanBobi for his [report and investigation](https://github.com/wwayne/react-tooltip/pull/211#issuecomment-287066189).
Thanks @wwayne for awesome `ReactTooltip` library.


MatanBobi:
Was planning on doing this, already forked the repo, Thanks and sorry.

huumanoid:
@MatanBobi, oh, I'm so sorry, this fix is really yours! Henceforth, let us know in advance, please! I'm sorry! I haven't noticed that you have created repository just yesterday...

MatanBobi:
@huumanoid its not your fault at all, It was a minor fix, definitely my bad.
Thanks for the fix!
It looks good.

Issue #137, opened 2016-07-16
Listen to body mouseover/mouseout rather than on individual elements?


mattzeunert:
It looks like the event listeners are added to each `[data-tip]` element when the ReactTooltip component mounts. That means I have to manually rebuild the bindings when the UI changes.

Would it be possible to add a mouseover listener to the body and then check if the event target element has a `data-tip` attribute?

wwayne:
It could be a solution, but not sure if it would have some other side effects, will make experiments later.

oyeanuj:
Also interested in this!

brafdlog:
This is a significant issue.
In the current implementation if an element is rendered after the componentDidMount of the ReactTooltip component, the tooltips on it wont work.
Calling rebuild is not a solution becuase I need to know to call rebuild anywhere there might be a tooltip, and rebuild is quite heavy - calling it on every render will eventually crash the browser tab.
@wwayne what do you think?

lanjingling0510:
Having the same issue. Dynamic elements have not come out, document.querySelectorAll not get data-tip elements, resulting in tooltip not come out.

PR #273, opened 2017-03-20
Feature: listen body events

huumanoid:
This PR related to feature request #137.
This feature is very suitable for me too, so I've tried to implement
it as safe as possible.

# DOM Event properties

In the DOM event object, there are three properties which I think are
interesting for this feature:
1. `currentTarget` - when listener is attached to the `<body>` element,
the `currentTarget` is equals to the `<body>`. That's why it's
inappropriate to call methods like `showTooltip` with such an event
object given.
2. `target` - this is the element which raised the event.
3. `path` - this is an actual path of the event's bubbling flow. We could
traverse `path` to find the element with `[data-tip]` set, even if the
`target` of the event doesn't have the `data-tip` attribute.

# Proposed solution

There are several goals that I pursued:
1. Don't change the existing code as much as possible.
2. Reuse the existing code as much as possible.

So, this PR introduces new mode for the `ReactTooltip`: `bodyMode`
When `ReactTooltip` works in body mode, it attaches listeners to a `<body>`
instead of each individual DOM element.
All events are captured in the "bubbling" phase.

## How to use

To activate the "body mode", you need to set `bodyMode` property:
```jsx
<ReactTooltip bodyMode />
```

## Basic behavior (no custom events)

For every basic event: `mouseenter`, `mousemove` and `mouseleave`,
the proposed implementation attaches listener to the `<body>` for events `mouseover`,
`mousemove` and `mouseout` respectively. 
These listeners has the following algorithm:
1. Check, if `e.target` has `data-tip` attribute and has proper `data-for`.
2. If yes, make proxy of the event object, with `target` assigned to `currentTarget`.
3. Then, call the corresponding function: `showTooltip`, `updateTooltip`
or `hideTooltip` with proxy event given.

`mousemove` is attached, no matter what the `props.effect` is. But
the effect is still handled correctly. There may be a performance issue
for someone who uses only `solid` tooltips.

## Custom events

To handle custom events, corresponding event listeners are attached to
the `<body>`. Event listener attached for every known event type.
The list of known event types consists of:
1. `props.event` (`props.eventOff`)
2. `data-event` (`data-eventOff`) of every element of the `targetArray`
at the time of call `bindBodyListener`
3. `props.possibleCustomEvents` (`props.possibleCustomEventsOff`). This
is the useful suggestion for tooltip, because the `targetArray` may not
contain all the required `data-event` values at the mount time.

## Unbinding

All the event listeners saved into `this.bodyModeListeners`.
So, they are safely detached on unmount/rebuild.

# Known issues

* Nested elements.
When an element associated with the `ReactTooltip` has no nested elements,
everything works well. Otherwise, tooltip would disappear as soon as mouse
pointer reaches nested element. This issue can be fixed by lookup of proper element in `e.path` array.

* No "capture" mode.
For current implementation, `isCapture` is ignored. For some cases, I guess, this issue is related to the previous one, and the "capturing" semantics can be implemented by selecting the appropriate lookup direction.

### DIFF
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -98,7 +98,7 @@ export default function (target) {
 
     const listeners = this.bodyModeListeners
     for (const event in listeners) {
-      body.removeEventListener(event, listeners(event))
+      body.removeEventListener(event, listeners[event])
     }
   }
 }
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -59,7 +59,8 @@ export default function (target) {
 
   target.prototype.bindBodyListener = function (targetArray) {
     const { id } = this.props
-    const { event, eventOff } = this.state
+    const { event, eventOff, possibleCustomEvents,
+      possibleCustomEventsOff } = this.state
     const body = getBody()
 
     const customEvents = findCustomEvents(targetArray, 'data-event')
@@ -67,8 +68,8 @@ export default function (target) {
 
     if (event != null) customEvents[event] = true
     if (eventOff != null) customEventsOff[eventOff] = true
-    for (const event of this.state.possibleCustomEvents) customEvents[event] = true
-    for (const event of this.state.possibleCustomEventsOff) customEventsOff[event] = true
+    possibleCustomEvents.split(' ').forEach(event => customEvents[event] = true)
+    possibleCustomEventsOff.split(' ').forEach(event => customEventsOff[event] = true)
 
     this.unbindBodyListener(body)
 
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -64,6 +64,8 @@ export default function (target) {
 
     if (event != null) customEvents[event] = true
     if (eventOff != null) customEventsOff[eventOff] = true
+    for (const event of this.state.possibleCustomEvents) customEvents[event] = true
+    for (const event of this.state.possibleCustomEventsOff) customEventsOff[event] = true
 
     this.unbindBodyListener(body)
 
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -54,7 +54,7 @@ const getBody = () => document.getElementsByTagName('body')[0]
 
 export default function (target) {
   target.prototype.isBodyMode = function () {
-    return this.props.bodyMode
+    return !!this.props.bodyMode
   }
 
   target.prototype.bindBodyListener = function (targetArray) {
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -57,10 +57,9 @@ export default function (target) {
     return this.props.bodyMode
   }
 
-  target.prototype.bindBodyListener = function () {
+  target.prototype.bindBodyListener = function (targetArray) {
     const { id } = this.props
     const { event, eventOff } = this.state
-    const targetArray = this.getTargetArray(id)
     const body = getBody()
 
     const customEvents = findCustomEvents(targetArray, 'data-event')
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -1,0 +1,57 @@
+/**
+ * Util method to get effect
+ */
+
+export default function (target) {
+  const clone = (e) => {
+    const copy = {}
+    for (const key in e) {
+      copy[key] = e[key]
+    }
+    return copy
+  }
+
+  const bodyListener = function (callback, respectEffect, e) {
+    const {id} = this.props
+
+    const tip = e.target.dataset.tip
+    const _for = e.target.dataset.for
+
+    const target = e.target
+
+    if (tip != null
+      && (!respectEffect || this.getEffect(target) === 'float')
+      && (
+      (id == null && _for == null)
+        || (id != null && _for === id)
+    )) {
+      const x = clone(e)
+      x.currentTarget = target
+      callback(x)
+    }
+  }
+
+  target.prototype.bindBodyListener = function () {
+    const body = document.getElementsByTagName('body')[0]
+
+    const listeners = this.bodyModeListeners = {
+      'mouseover': bodyListener.bind(this, this.showTooltip, false),
+      'mousemove': bodyListener.bind(this, this.updateTooltip, true),
+      'mouseout': bodyListener.bind(this, this.hideTooltip, false)
+    }
+    for (const event in listeners) {
+      body.addEventListener(event, listeners[event])
+    }
+  }
+
+  target.prototype.unbindBodyListener = function () {
+    const body = document.getElementsByTagName('body')[0]
+
+    const listeners = this.bodyModeListeners
+    for (const event in listeners) {
+      body.removeEventListener(event, listeners(event))
+    }
+  }
+}
+
+
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -3,61 +3,65 @@
  */
 import {checkStatus} from './customEvent'
 
+const makeProxy = (e) => {
+  const proxy = {}
+  for (const key in e) {
+    if (typeof e[key] === 'function') {
+      proxy[key] = e[key].bind(e)
+    } else {
+      proxy[key] = e[key]
+    }
+  }
+  return proxy
+}
+
+const bodyListener = function (callback, options, e) {
+  const {respectEffect = false, customEvent = false} = options
+  const {id} = this.props
+
+  const tip = e.target.dataset.tip
+  const _for = e.target.dataset.for
+
+  const target = e.target
+  if (this.isCustomEvent(target) && !customEvent) {
+    return
+  }
+
+  const isTargetBelongsToTooltip =
+    (id == null && _for == null) || (id != null && _for === id)
+
+  if (tip != null &&
+    (!respectEffect || this.getEffect(target) === 'float') &&
+    isTargetBelongsToTooltip
+  ) {
+    const proxy = makeProxy(e)
+    proxy.currentTarget = target
+    callback(proxy)
+  }
+}
+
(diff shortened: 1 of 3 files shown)


huumanoid:
And of course, I've forgot about something important

# Testing

Everyone who wants to test this feature: you may install dist from [this branch](https://github.com/huumanoid/react-tooltip/tree/feature-listen-body-dist)

```sh
npm install github:huumanoid/react-tooltip#feature-listen-body-dist
```

# Thanks

Thanks @mattzeunert for this amazing idea, it rocks! Thanks everyone who is interested in #137, it makes me feel I'm doing something important.
Thanks everyone who will review and test my changes, this feature is not as simple as it could be, so your help is important for me and for this library as well.

wwayne:
Yes, this is very important, it could solve a lot of bugs that caused by my previous design... But I don't time to check the PR recently, maybe this weekend?

huumanoid:
@wwayne I don't think this PR is easy to review, so of course we can wait for it. 
There is an option that my solution is not bold enough to solve some bugs.
I think we should seriously discuss this PR a lot before merging it.

Thank you for fast reply! Looking forward to feedback from you.

wwayne:
@huumanoid BTW, do you want to take over the project? I'm not able to spend too much energy on this open-source project nowadays.  I checked your github, I think you are reliable and very active.

It was a pretty good experience to maintain an open-source project that used by lots of people, I made couple mistakes and learnt a lot from it. Both you and the project would get benefit from it, how do you think?

If it's okay for you, could you tell me your NPM account so that you are able to do release, not sure if I need to add you into travis CI manually, I will try to figure it our.

stevenpietzsch:
Hello @huumanoid, I have tested your PR. Unfortunately it is not working for me. I used this markup:

```
import ReactTooltip from 'react-tooltip';
...
<span>
  <span data-tip='true' data-for={id}>
    Hover me
  </span>
  <ReactTooltip id={id} place='top' bodyMode>
    My tooltip
  </ReactTooltip>
</span>
```

``id`` is a uuid. 
I get the following errors in console after the page is mounted:

> bodyMode.js:63 Uncaught TypeError: listeners is not a function
    at ReactTooltip.target.unbindBodyListener (bodyMode.js:63)
    at ReactTooltip.unbindListener (index.js:257)
    at ReactTooltip.componentWillUnmount (index.js:169)
    at ReactCompositeComponentWrapper.unmountComponent (ReactCompositeComponent.js:243)
    at Object.unmountComponent (ReactReconciler.js:52)
    at Object.unmountChildren (ReactChildReconciler.js:115)
    at ReactDOMComponent.unmountChildren (ReactMultiChild.js:398)
    at ReactDOMComponent.unmountComponent (ReactDOMComponent.js:911)
    at Object.unmountComponent (ReactReconciler.js:52)
    at Object.updateChildren (ReactChildReconciler.js:88)

What am I doing wrong?

huumanoid:
Hello @stevenpietzsch and thank you for the feedback, really appreciate it!

Hell no, I AM doing something wrong.
I'm very sorry but the dist has been broken because of silly error. Sorry for wasting your time.

I have fixed it and tested in all my own use cases. I hope it would work for you, too.
At least, your example seems to be free from issues of "known issues" section.

I have published the dist again at the same [branch](npm install github:huumanoid/react-tooltip#feature-listen-body-dist). Please, check it! Thank you.

huumanoid:
@wwayne I'm really proud that someone like you offers me to maintain such a project.
Yes, I've already get some benefits from maintaining `ReactTooltip`, and I really like it!

I'm brave enough to accept! Thank you for your trust.
However, I feel lonely when I think about maintaining `ReactTooltip` without you.

My npm is:
https://www.npmjs.com/~hmnid

wwayne:
@huumanoid Great to hear that, from my experience, people from the community will help you a lot and I will try my best to help you. 

Would it be better transferring the repo to your account? I just afraid it will affect some links like example page, CI, Let me know if you are ready to accept the repo into your Github account.

Could you give me your email? Just for your reference, I will let you know how I do release, etc. I will do this late of today, thanks again :)

stevenpietzsch:
Hi @huumanoid, thanks for your response! Your dist is working now. Unfortunately, your fix is apparently not that I was looking for. My issue is the following: We are working with [react-datagrid](https://github.com/zippytech/react-datagrid) and [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap)-Overlays. Because of some issues in this repo we intend to replace it by wwayne's react-tooltip which is very neat IMHO. But we encounted two issues with react-tooltip:

1. The tooltip is not correctly positioned in IE 11 if CSS3 transform-styles are applied. Maybe this could be fixed by #222. After applying this PR, I have to redefine my popover styles, but that will be ok.
2. When using react-tooltip, the tooltip is rendered hidden behind the react-datagrid header. The react-datagrid is a scrolling div container with fixed headers. My hope was, that your PR #273 would fix this issue, but it doesn't. Wasn't this the issue your we're targeting? I thought, after your PR the tooltip will be added to the document.body which is done in react-bootstrap. Please compare the following to images to comprehend my issue.

**react-bootstrap-tooltip**
react_bootstrap_tooltip_in_data_grid

**wwayne-tooltip with applied #273** 
wwayne_tooltip_in_data_grid

stevenpietzsch:
Ok, I am one step further. I need to render the tooltip outside of the react-datagrid. Inside each cell I need to render just the element that triggers the tooltip. 

**trigger element inside cell**
```
<span data-tip='true'>Hover me</span>
```

**tooltip outside react-datagrid**
```
<ReactTooltip bodyMode>
  My Tooltip
</ReactTooltip>
```

After that, the tooltip is correctly positioned, even in IE 11 - without #222. Great work, man! :blush:
But one last question: Is it now possible to update the tooltip's content? Currently it's "My Tooltip". I want to render the ``<ReactTootip/>`` just once on the page outside of the react-datagrid and want to render different tooltip contents coming from individual datagrid cell contents, e.g. "Tooltip for cell 1", "Tooltip for cell 2" etc.
wwayne_tooltip_ouside_datagrid

huumanoid:
@stevenpietzsch hello and I'm glad to hear this patch is works for you!
I'm sorry for the delay!

# Important (the answer)
```jsx
<ReactTooltip bodyMode />

<div data-tip="my custom tip one" />
<div data-tip="my custom tip two" />
```

### chat

It seems like `react-datagrid` is not the only grid which affected by the problem you described.
There are actually a workaround to make the original `ReactTooltip` work in `react-datagrid`:
```js
class Derived extends ReactDatagrid {
  componentDidUpdate() { ReactTooltip.rebuild() } // to handle virtual scrolling
  componentDidMount() { ReactTooltip.rebuild() }
}
```
I can't wait I migrate from this solution. `ReactTooltip.rebuild` + virtual scrolling doesn't seem good.

huumanoid:
@wwayne thanks! I'm collaborating `react-tooltip` at npm now!
Thank you for your parting words, I'm braver than ever now.G
I don't think you should transfer this repository to me.

My email is huumanoid@hotmail.com. Looking forward to reference from you!

stevenpietzsch:
@huumanoid Thank's for your help. Your tip was good - for plain strings. I have attached again two screenshots. The first one (``renderSimpleTooltip()``) works quite nice, in the second example (``renderAdvancedTooltip()``) I wanted to render the same as React component. Do you have any clues how to render React components like if I would have used the markup inside of ``<ReactTooltip/>``.
```
function getTooltip(text) {
  return `
    <div class='test' style='color: red'>${text}</div>
  `;
}

export function renderSimpleTooltip() {
  return (
    <span data-tip={getTooltip('My Tooltip')} data-html={true}>Hover me</span>
  );
}

export function renderMyTooltip(id) {
  return (
    <ReactTooltip bodyMode/>
  );
}
```

tooltip

```
function getAdvancedTooltip(text) {
  return (
    <div className='test' style={{color: 'red'}}>${text}</div>
  );
}

export function renderAdvancedTooltip() {
  return (
    <span data-tip={getAdvancedTooltip('My Tooltip')} data-html={true}>Hover me</span>
  );
}

export function renderMyTooltip(id) {
  return (
    <ReactTooltip bodyMode/>
  );
}

```
tooltip2

stevenpietzsch:
IMHO, you got 2 options:

1. Append the ``<ReactTooltip/>`` node to the document.body on ``mouseover``. On ``mouseout`` remove it from document.body. This is done in the ``react-overlays`` package:

[Position.js](https://github.com/react-bootstrap/react-overlays/blob/master/src/Position.js)
[Overlay.js](https://github.com/react-bootstrap/react-overlays/blob/master/src/Overlay.js)

Thus, all rendering quirks in scrollable containers should be prohibited.
```
<Overlay 
  container={document.body}
  target={() => ReactDOM.findDOMNode(this.refs.triggerElem)}>
  <ComplexComponentToRenderInsideTooltip/>
</Overlay>
```

2. Use [RxJS](https://github.com/Reactive-Extensions/RxJS) and listen to events coming from other components to render an updated string or complex markup. But eventually, this might be too heavyweight for ``react-tooltip``.

huumanoid:
@stevenpietzsch I really love your explanation! It seems that it's a significant feature. And, as you said, it resolves a lot of issues.

tyler1314371:
Hi, is this PR still active? A lot of us are very interested in this feature

cvburgess:
@huumanoid is there anything I can do to help move this along?

ThatBrianDude:
+1. still waiting for this!

infiman:
+1. Really helpful PR. Waiting.

aronhelser:
This PR can't be merged in it's current state because of conflicts. I haven't had the time to resolve and test. If someone wants to look, that'd be great. I think @huumanoid isn't working with react currently.

artdevgame:
@huumanoid Thanks for your contribution, is there any way you could help get this work merged into master? There's a few of us that could really use your hard work.

artdevgame:
Fixed conflict in #526

PR #287, opened 2017-04-13
Use prop-types package instead of React.PropTypes

ssilve1989:
In response to warnings appearing when using React 15.5+

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "react-dom": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "classnames": "^2.2.0"
+    "classnames": "^2.2.0",
+    "prop-types": "^15.5.8"
   },
   "engines": {
     "node": ">=4.2.1"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,8 @@
     "react-dom": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "classnames": "^2.2.0"
+    "classnames": "^2.2.0",
+    "prop-types": "^15.5.8"
```
audiolion: this should be under `devDependencies`

ssilve1989:
Actually its suggested to place it in dependencies.

A quote from the library page.
```
For libraries, we also recommend leaving it in dependencies:

  "dependencies": {
    "prop-types": "^15.5.0" 
  },
  "peerDependencies": {
    "react": "^15.5.0" 
  }
```

https://www.npmjs.com/package/prop-types

wwayne:
Yeah thanks @ssilve1989 and @audiolion , merging

PR #291, opened 2017-04-20
Add ability to specify custom DOM element for `scrollHide` scroll events

chriscain:
Just started using `react-tooltip` this week, really awesome library - thank you guys!

Currently, `scrollHide` just looks for scrollEvents on the `window`. This works well 99% of the time, but if a page has custom scroll containers, the `window` never actually sees any of those scroll events, as they're not fired on the `window`.

This PR just adds a way to plug in a custom selector to replace the window - I don't believe it fixes every problem of https://github.com/wwayne/react-tooltip/issues/248, but it _should_ act as an easy workaround for situations like @philraj was having [here](https://github.com/wwayne/react-tooltip/issues/248#issuecomment-288132725).

I haven't been following the project closely, so I hope this is the proper way to contribute. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ className	|   data-class  |  String  |   | extra custom class, can use !importan
  afterHide | null | Func | () => {} | Function that will be called after tooltip hide
  disable | data-tip-disable | Bool | true, false | Disable the tooltip behaviour, default is false
  scrollHide | data-scroll-hide | Bool | true, false | Hide the tooltip when scrolling, default is true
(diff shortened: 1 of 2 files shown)


gaohailang:
Useful PR

gregoryfabry:
For anyone running into this issue, setting the isCapture prop will allow the window scroll listener to catch scroll events on a div when using overflow: scroll.

aronhelser:
Confirmed that setting isCapture fixes the issue. Perhaps we should make isCapture the default?

aronhelser:
Handled by isCapture.

PR #293, opened 2017-04-26
Custom Type for Unstyled Tooltip

wylie:
making a tooltip that has no presentation styles that need to be overridden with `!important`

- moving some presentation styles so `data-type="custom"` can be used
- adding `custom` to `tooltipClass` so we return no `type-className` class
- updating the readme to reflect the newness


danqing:
Would be great to see this merged!

src/index.scss
```diff
@@ -1,5 +1,8 @@
 @mixin type($background-color) {
   background-color: $background-color;
+  border-radius: 3px;
+  font-size: 13px;
+  padding: 8px 21px;
```
aronhelser: I see what you want to do - but adding them to something named 'background-color' seems misleading. Maybe we can define another mixing type, like base-sizing or something?

src/index.js
```diff
@@ -439,7 +439,8 @@ class ReactTooltip extends Component {
       {'type-warning': this.state.type === 'warning'},
       {'type-error': this.state.type === 'error'},
       {'type-info': this.state.type === 'info'},
-      {'type-light': this.state.type === 'light'}
+      {'type-light': this.state.type === 'light'},
+      {'type-custom': !this.state.type}
```
aronhelser: so 'custom' ends up being the same as not specifying a type, correct?

To merge this, I'd like to see an example added to illustrate how to use it.

PR #301, opened 2017-05-19
insert css as first to allow easy css styling without important

roblan:
As in title.
It will lower it's specificity to minimum. So no !important will be needed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -407,11 +407,12 @@ class ReactTooltip extends Component {
    * in this way we can insert default css
    */
   setStyleHeader () {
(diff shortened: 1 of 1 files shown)


whitecrownclown:
When could we merge this? I think it's a good change :)

pedroabreu:
+1 for this change. I've been increasing the specificity to change the default styles. Ideally I wouldn't have to do that (same as using the `!important`)

xrado:
+1

roblan:
@huumanoid what do You think?

uselessdev:
hi @aronhelser this fix will be publish?

aronhelser:
Yes, in the next day or two - there are a few other PR's I'd like to get in
as well.

Issue #297, opened 2017-05-05
`undefined is not a function` on Chrome 38


danieleleoni:
Hi,
I was updating react-tooltip from 3.2.9 to 3.3.0, but I realised there is an error on trackRemoval.js on this row, even if tooltip is working fine
```
for (var _iterator2 = mutation.removedNodes[Symbol.iterator](), _step2; !(_iteratorNormalCompletion2 = (_step2 = _iterator2.next()).done); _iteratorNormalCompletion2 = true) {
```
<img width="1280" alt="screen shot 2017-05-05 at 17 37 16" src="https://cloud.githubusercontent.com/assets/20039309/25752757/8520db6e-31b9-11e7-8d56-f2b35995b467.png">

Problem is not present at all on 3.2.9 or into a recent version of Chrome

Thanks!

OriR:
@huumanoid 
This error appears in Edge as well, could you please take a look at this?

Alfafc:
Same here, last update throw those errors

icodeforyou:
Safari 9 throws this error aswell.

avivshafir:
+1

nik-lampe:
+1

Rogger794:
Can someone check if the error already exist?

PR #302, opened 2017-05-30
Changed for loop for ES5

andresalves:
Browser EDGE broken.

Error: Object doesn't support property or method 'ToString'


OriR:
@andresalves 
Do you know if this by any chance solves #297 ?
The error is different but looks like the same piece of code.

andresalves:
@OriR  Yes. Basically it was the same as I did.
Are you going to do the fix?

OriR:
Unfortunately no, I ran into this PR by mistake after I noticed there was already an issue about this.
I really hope this gets merged soon!
The weird thing is that I'm using [babel transform-es2015-for-of](https://babeljs.io/docs/plugins/transform-es2015-for-of/) in my project so it should work but for some reason it doesn't...

pemasora:
Need this changes asap. Any prevision?

PR #309, opened 2017-06-19
Problem with solid effect for touchscreen devices

BysahaOleh:
- Given tooltip issue
- When I open the tooltip and make scrolling of the page
- Then Tooltip disappears and it's impossible to evoke it until focus remains on the element

**Problem arises only when effect = solid on touchscreen devices**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -182,7 +182,7 @@ class ReactTooltip extends Component {
         return
       }
 
-      target.addEventListener('mouseenter', this.showTooltip, isCaptureMode)
+      target.addEventListener('mousemove', this.showTooltip, isCaptureMode)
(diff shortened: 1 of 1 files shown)


aronhelser:
This change seems risky for non-touchscreen usage - now we are calling `showTooltip` on every mouse move? Maybe we can make it specific to the case that is a problem?

BysahaOleh:
Hi @aronhelser! I don't remember what problem I had. I fixed it this way. Maybe it's realy specific case for me. So if You think it's risky, just close this PR.

aronhelser:
OK, thanks. I'll close and if someone runs into it they can reopen.

PR #310, opened 2017-06-19
Update position with content

stefanhayden:
When length of content changes in an interval the position never updates when the mouse is still. This causes the positioning of the tooltip to be in the wrong place.

a simple call to `this.updatePosition()` at the end of `this.intervalUpdateContent` always keeps the tooltip in the correct position.


aronhelser:
I like it! I checked it out, and I'm fine with modifying the example.

stefanhayden:
thanks! you rock!

stefanhayden:
I'll try and look in to it.

PR #360, opened 2018-01-03
Compute or enrich tip content

austil:
Hi, for a specific use case I needed to compute the tip content at runtime based on the data-tip's attribute value, so here we go.

PS :  I may have interests in maintaining the project 😉


aronhelser:
This makes perfect sense to me. Then each target could actually just have `data-tip={id}` that lets the `getContent(dataTip)` method identify which element it is.

BTW, @austil I would be happy for any help you would care to give in maintaining, if you are interested.

austil:
Conflicts on this pull request are weird, I should probably submit a new one (changes are fairly simple), ready to merge.

@aronhelser I will be happy to help. What are your thought on the build system (compared to stuff like webpack) and "test galery" (compared to a shiny storybook) of this project ?

austil:
Some refactoring was made on the part concerned by my changes so I had to do a new commit (while deleting the previous ones with a dirty head reset followed by a force push).
It is now ready to merge :wink:

phileveritt:
thank you both for working on this and getting it merged - it's a great addition.

PR #366, opened 2018-02-13
Make getPosition() calculate offsets correctly for svg elements

P0lip:
Current implementation of getPosition is broken for children of svg elements.
Their clientHeight and clientWidth values are usually (or always, not sure, though) equal to 0, therefore calculated offsets are wrong.
This may be a common issue for tooltips whose tips are some sort of svg child elements, like polygon, rect, line, circle etc. In such cases, their placement is usually wrong.

### DIFF
--- a/src/utils/getPosition.js
+++ b/src/utils/getPosition.js
@@ -15,10 +15,18 @@
  * - `position` {OBject} {left: {Number}, top: {Number}}
  */
 export default function (e, target, node, place, effect, offset) {
-  const tipWidth = node.clientWidth
-  const tipHeight = node.clientHeight
+  const {
+    width: tipWidth,
+    height: tipHeight
+  } = getDimensions(node)
+
+  const {
+    width: targetWidth,
+    height: targetHeight
+  } = getDimensions(target)
+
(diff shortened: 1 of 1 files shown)


austil:
Hey, I runned into the exact same problem at work (tooltip on svg elements). 
I cherry picked your commit and it work perfectly. Thanks a lot !

austil:
Check this https://bugzilla.mozilla.org/show_bug.cgi?id=874811

tldr : 

> SVG elements don't have a CSS layout box  
> The clientTop, clientLeft, clientWidth, and clientHeight attributes must return zero if the element does not have any associated CSS layout box

> use getBoundingClientRect

aronhelser:
Looking at https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth, it seems `getBoundingClientRect()` will return fractional values. Otherwise it seems equivalent, and works for SVG. Have you seen any ill-effects because of fractional values?

aronhelser:
Oops, sorry, I missed your comment before I merged. Feel free to submit another PR, I'll get it right in.
If you start your commit message like:
> fix(examples): add SVG example
then I can test auto-publish of the examples 😄

PR #369, opened 2018-03-07
fix(index.js): add missing argument so tooltip hides.

aronhelser:
If getContent returns '', the tooltip should hide. A missing
argument caused this not to work for dynamic content.


aronhelser:
In my case, I was trying to display content based on position in an image:
`        <ReactTooltip
          id="overTime"
          getContent={[this.getTooltip, 50]}
        />
        Element rendering
`

With this fix, if `this.getTooltip()` returns '' (empty string), the tooltip hides.

PR #376, opened 2018-04-19
Add displayName

CLL80:
Using `react@16.3` and `enzyme@3.3.x` to test, the name of the component is now `<t>` when previously in `react@16.2` it was `<ReactTooltip>`. Setting the displayName property on the component class ensures the name will remain as `ReactTooltip`. For example, using enzyme:
```js
expect(ShallowInstance.children().last().name()).to.equal('ReactTooltip')
```


aronhelser:
Awesome, thanks! 

If you would like your fix to trigger a release, reformat the commit message commitizen-style, like this:

`fix(index): Add displayName`

CLL80:
I'm glad to contribute. Thanks for the tooltip. I hadn't seen commitizen before. Do you want me to close and reopen with the updated commit message?

CLL80:
@aronhelser, I found that this is actually due to changes in webpack 4, and the name is retained when using `development` mode. For example, in `karma.config.js`:
```js
module.exports = {
  ...
  webpack: {
    ...
    mode: 'development',
    ...
  }
}
```

Anyway, I still think it isn't a bad idea to define your `displayName`, in case you'd still like to merge this PR.

aronhelser:
Thanks!

Issue #312, opened 2017-06-22
Uncaught TypeError: Cannot read property 'getAttribute' of null


Springbach:
bundle.js:13915 Uncaught TypeError: Cannot read property 'getAttribute' of nulltarget.isCapture @ bundle.js:13915addScrollListener @ bundle.js:13464(anonymous function) @ bundle.js:13348renderComponent @ bundle.js:2182rerender @ bundle.js:1909

source map

exports.default = function (target) {
	  target.prototype.isCapture = function (currentTarget) {
	    var dataIsCapture = currentTarget.getAttribute('data-iscapture');
	    return dataIsCapture && dataIsCapture === 'true' || this.props.isCapture || false;
	  };

please help

samsiu:
I believe this has to do with react nullifying the event object after setState. see: https://github.com/facebook/react/issues/2850
The recommended way is to cache the e.currentTarget to component's state, I think.

alex-shamshurin:
The same issue is here. Can we replace `currentTarget` with `target` as temporary workaround?

occamsraiser:
I'm seeing this problem as well.

alex-shamshurin:
@occamsraiser Try this fork https://www.npmjs.com/package/react-tooltip-currenttarget

occamsraiser:
@alex-shamshurin I tried it out -- I get a `Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null` from getPosition.js in this fork

alex-shamshurin:
@occamsraiser , I use this fork in production and get no errors.  My fork fixed only one issue here

```
  /**
   * Add scroll eventlistener when tooltip show
   * automatically hide the tooltip when scrolling
   */
  addScrollListener (e) {
    const isCaptureMode = this.isCapture(e.target)
    window.addEventListener('scroll', this.hideTooltip, isCaptureMode)
  }

```

not other issues related with the topic, your problem is related with the same reason (react lose currentTarget after setState callback). I'll try to fix it on fork.

alex-shamshurin:
@occamsraiser I have updated to 3.3.4 and then got the error. So 3.3.4 is broken, I will fix it, but 3.3.3 works normally

alex-shamshurin:
I've just released 3.3.5 https://www.npmjs.com/package/react-tooltip-currenttarget and the issue is fixed, try it and please report and I make a PR to main package

occamsraiser:
[...]

PR #389, opened 2018-06-04
fix(isCapture): guard use of currentTarget

aronhelser:
currentTarget was undefined, save it's value using
state before using in scroll listener.

@alex-shamshurin, @renanborgez does this fix issue #312 for you? I've tried to simplify the fix as much as possible.


renanborgez:
@aronhelser 
i downloaded you branch, builded and tested

....and it works like a charm!!!!

aronhelser:
Good enough for me 😄

PR #391, opened 2018-06-06
fix(positioning): make sure tooltip is oriented correctly when close to edge

### LINKED ISSUE #371: Incorrect tooltip positioning in corner cases
Tooltip won't display correctly when close to top-left and top-right corner of the browser window.

How to reproduce:
- align the hoverable area to be near the top-right corner of the window
- position mouse cursor above the hoverable area
- move cursor down until it is inside of the hoverable area
- tooltip will be displayed with incorrect position

Following picture was taken from http://wwayne.com/react-tooltip/
image

hassanbot:
We now check different tooltip orientations to make sure the best one is chosen. There are 2 cases that lead to reorientation:
1.  If the desired orientation (`desiredPlace`) is completely inside the client window, and the current orientation (`place`) is different from the desired one (then **reorient to the desired orientation**)

2.  If neither the desired orientation (`desiredPlace`) nor the current one (`place`) is inside the client window, but there are other orientations that are (then **reorient to an arbitrary orientation that is inside**)

All other cases will keep the orientation as it was before.

Fixes #371

### DIFF
--- a/src/utils/getPosition.js
+++ b/src/utils/getPosition.js
@@ -52,139 +52,56 @@ export default function (e, target, node, place, desiredPlace, effect, offset) {
     return mouseY + offset_Y + extraOffset_Y
   }
 
-  // Judge if the tooltip has over the window(screen)
-  const outsideVertical = () => {
-    let result = false
-    let newPlace
-    if (getTipOffsetTop('left') < 0 &&
-      getTipOffsetBottom('left') <= windowHeight &&
-      getTipOffsetBottom('bottom') <= windowHeight) {
-      result = true
-      newPlace = 'bottom'
-    } else if (getTipOffsetBottom('left') > windowHeight &&
-      getTipOffsetTop('left') >= 0 &&
-      getTipOffsetTop('top') >= 0) {
-      result = true
-      newPlace = 'top'
+  //
+  // Functions to test whether the tooltip's sides are inside 
+  // the client window for a given orientation p
+  //
+  //  _____________
+  // |             | <-- Right side
+  // | p = 'left'  |\
+  // |             |/  |\
+  // |_____________|   |_\  <-- Mouse
+  //      / \           |
+  //       |
+  //       |
+  //  Bottom side
+  //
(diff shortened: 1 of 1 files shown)


src/utils/getPosition.js
```diff
@@ -52,139 +52,55 @@ export default function (e, target, node, place, desiredPlace, effect, offset) {
...
+  let outsideLeft = p => getTipOffsetLeft(p) < 0
+  let outsideRight = p => getTipOffsetRight(p) > windowWidth
+  let outsideTop = p => getTipOffsetTop(p) < 0
+  let outsideBottom = p => getTipOffsetBottom(p) > windowHeight
+
+  // Check whether the tooltip with orientation p is completely inside the client window
+  let outside = p => outsideLeft(p) || outsideRight(p) || outsideTop(p) || outsideBottom(p)
+  let inside = p => !outside(p)
+
+  let placesList = ['left', 'right', 'top', 'bottom']
```
aronhelser: I think we might want 'top' and 'bottom' to have priority - they will if you list them first, right?

I really like how you've simplified the code - it's much clearer. I still need to try it out (or someone does)

aronhelser: Oh, and please format the first line of you commit message for semantic versioning:

'fix(positioning): make sure ...'

If you do, travis will do an automatic release.

hassanbot: I've changed the pull request title now, is that enough?

You're right about the orientation priority, it's changed with 6f01ed8dfcb25823daca3688563454c9589f971c

aronhelser:
I was able to test, and it looks good. It also fixes the problem with entering from the top. Merging!

aronhelser:
I'm sorry, I don't understand. Can you provide an example or steps to reproduce the problem?

aronhelser:
And potentially file an issue...

TotallWAR:
image

As you can see, reposition works perfectly, however both of pseudo elements (for both posotions) are shown

TotallWAR:
image

It's an example without reposition, when it works default.

TotallWAR:
Any ideas?

Issue #388, opened 2018-05-31
Rerender at initial position


an4ger:
1. Create tooltip with delayShow={300} delayHide={300} place="top" options.
2. Move cursor to element with tooltip and wait for tooltip, **it should appear not at the top position**.
3. Quick move cursor from element with tooltip and back.

Result: tooltip try to render with bottom arrow as at top place (initial option) and after that appear at right position. So, we have some blinking effect. 
If set delay to 5 seconds we have the same effect.

https://codesandbox.io/s/j70q5w7p1w
![tooltip](https://user-images.githubusercontent.com/5645926/40783471-2a658768-64eb-11e8-8f39-7e03ad1ed45e.gif)

smakosh:
I'm having the same issue when the tooltip should appear near the corner of the window

an4ger:
Here some info after investigation:

_mouseenter_ event trigger this function:

https://github.com/wwayne/react-tooltip/blob/c44cc2d0aa89b7e6268a61a277326be6903b00ae/src/index.js#L256

and than we call 

https://github.com/wwayne/react-tooltip/blob/c44cc2d0aa89b7e6268a61a277326be6903b00ae/src/index.js#L404

here we have some problem:
` if (result.isNewState) {
      // Switch to reverse placement
      return this.setState(result.newState, () => {
        this.updatePosition()
      })
    }`

in my case function updatePosition triggers 2 times: 
- first time result.isNewState = true, because in showTooltip function we set desiredPlace in state every time on mouseenter and that desiredPlace not equal current available position;
- second time result.isNewState = false and all good;

So, it will be better to ignore first call of updatePosition  function.

One of solution is to do some caching and try to set desiredPlace not from attribute but from last available place:
https://github.com/wwayne/react-tooltip/blob/c44cc2d0aa89b7e6268a61a277326be6903b00ae/src/index.js#L286

lets try to change it this way:

`this.state.place || desiredPlace: e.currentTarget.getAttribute('data-place') || this.props.place || 'top',`

Is it good idea? Any suggestions?

P0lip:
Well, I didn't check it myself, yet I suppose you are right and `updatePosition` is called more than once.

The trick with `desiredPlace` may be a bit troublesome and may introduce new bugs, as it doesn't take into account any external changes that could affect the position of tooltip in the meantime.
By external changes, I mean window resize, page scroll or any other layout changes.
[...]

PR #413, opened 2018-08-20
fix(index.js): Use correct orientation when mouse enters

an4ger:
getPosition has to be called before first setting the state (and thus rendering the tooltip) when
mouse enters, to take into account cases when the position is close to the client window edge

#388


aronhelser:
This seems reasonable to me, I just need to check it out. Do you notice any different behavior on the examples?

an4ger:
Didn't find any problem

aronhelser:
Sorry @an4ger , I merged another PR that causes conflicts. Can you resolve them?

aronhelser:
It checks out, thanks!

PR #416, opened 2018-08-24
feat(mouseover): Add ability to hover on tooltip.

RobertGary1:
Provide optional delay of updating so if the mouse path into the tooltip crosses another target the contents don't change. This feature allows you to put menus in tooltips, etc.

issue 411


vireshshah:
@wwayne @RobertGary1 Any timeline to merge this PR? I think we need similar feature.

aronhelser:
This is a great idea, and I like the example, it works well. 

Unfortunately, on Windows Chrome, I get an error in the "Update tip content over time" and "Test Scrolling" sections:
Uncaught TypeError: Cannot set property 'intervalUpdateContent' of undefined
    at ReactTooltip.<anonymous> (index.js:330)
    at commitCallbacks (react-dom.development.js:6544)
....

@RobertGary1 or @vireshshah if you can look and resolve, we can get this merged - otherwise you'll have to wait for me to investigate, which unfortunately may take a while...

RobertGary1:
Oh, I'm sorry. I see the issue. Should be a quick fix.

vireshshah:
Awesome. Thanks a lot, @RobertGary1 @aronhelser. If everything looks good then can we merge it? As I mentioned, we're waiting eagerly for this fix. Thanks!

vireshshah:
Thanks a lot @aronhelser @RobertGary1

PR #445, opened 2018-10-15
feat(src/index.js): expose the original event to `afterShow` and `afterHide`

### LINKED ISSUE #406: afterShow and afterHide should not swallow the trigger event
for the event callbacks: afterShow and afterHide
should pass the original `event` in the callback
https://github.com/wwayne/react-tooltip/blob/master/src/index.js#L344
https://github.com/wwayne/react-tooltip/blob/master/src/index.js#L378
Into
`afterShow(e)` and `afterHide(e)`

I can create PR if people think this is valid

cjzcpsyx:
BREAKING CHANGE: `afterShow` and `afterHide` will now take in a parameter to handle the original
event

fix #406

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -385,7 +385,7 @@ class ReactTooltip extends React.Component {
           show: true
         }, () => {
           this.updatePosition()
-          if (isInvisible && afterShow) afterShow()
+          if (isInvisible && afterShow) afterShow(e)
         })
       }
     }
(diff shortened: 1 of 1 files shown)


cjzcpsyx:
@bluestreek18 
need maintainer involved to check

aronhelser:
This change looks fine. It is not a breaking change, since existing code will not assume an argument available to afterShow or afterHide. Can you remove the BREAKING CHANGE from the commit message? I don't want to bump the major version number automatically. Using feat(...): will bump the minor version number.

We need to document the change in readme.md.

cjzcpsyx:
fixed in https://github.com/wwayne/react-tooltip/pull/466

PR #461, opened 2018-12-13
Support for older versions of firefox

787acr:
'matches' function is not supported in older versions of firefox and currently this component only have a workaround implemented for old versions of internet explorer. I'm expanding that in order to support old versions of firefox as well.


aronhelser:
Great, thanks!

PR #464, opened 2019-01-01
fix(html): remove sanitize-html-react, reduce package size

### LINKED ISSUE #429: Opt out from sanitize-html safely? (package too big)
Hi, first of all many thanks for this package making life easier for us! 

I noticed that our *total* bundle size suddenly increased by *50%*, and after some detective work I pinpointed the issue to the html-sanitizer package added here: https://github.com/wwayne/react-tooltip/pull/422

This addition made react-tooltip 7(!) times bigger in size and unfortunately that is a dealbreaker for us. Now this was improved a bit by replacing it with sanitize-html-react  (https://github.com/wwayne/react-tooltip/pull/426) but it's still very big for a tooltip component (for our usecase).

I found this tool which gives a graph of the size changes, I don't know the accuracy but it should give an idea at least: https://bundlephobia.com/result?p=react-tooltip@3.8.0

I understand that this is for security purposes, but is there a way to *safely* opt out from this massive package, we don't really need to let users write the actual html,  but we do use some markup where we insert strings from users. 

Sorry for potentially stupid question since I don't know the code very well, but is there a way to avoid dangerousslySetInnerHtml and simply render children instead? If that removes the need for sanitizing.

Thanks

aronhelser:
Remove sanitize html, to reduce our package size. Document
the need for sanitization in the readme.

Fixes #429 

CC: @NaNdreas @sgiacinto @wmertens


wmertens:
LGTM :+1:

Issue #430, opened 2018-09-14
lots of findDOMNode invariant warnings since 3.8.x release


occamsraiser:
Hi folks. We've been using this library to power our website tooltips in React for about two years. We use Sentry.io to track front end errors experienced by visitors, and I've noticed a huge uptick in findDOMNode Invariant Violation warnings since we deployed 3.8.x.  Can anyone else corroborate? We use findDOMNode in two spots in our codebase and I'm seeing these warnings from most of our pages that have these tooltips on them.

The specific error is "findDOMNode was called on an unmounted component"

stahlmanDesign:
If you use strict mode, React warns that findDOMNode is deprecated. It is used in react-tooltip and would be good if it were refactored to use the proposed solution of adding "a ref directly to the element you want to reference" — I haven't looked into exactly what to do to fix this:
`
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of ReactTooltip which is inside StrictMode. Instead, add a ref directly to the element you want to reference.
`

stahlmanDesign:
Relates to #331 describing possible solution

aronhelser:
If you want to try fixing react-tooltip, I'll gladly accept a PR!

nadineshmo:
Any update on this PR? I'm also getting the issue "findDOMNode was called on an unmounted component" raised by the first commenter. Not sure if this PR will fix this, or just the strict mode deprecation warning?

Andsbf:
I wonder the same @nadineshmo, **apparently** it only fix the strict mode thing.

Andsbf:
@thanks for merging it in @aronhelser , will a new tag be released with this change?

oudia15:
I am still having this issue, Has a new tag been released with the fix ?

PR #475, opened 2019-03-07
Refactor(findDOMNode, string ref): replace deprecated findDOMNode and string ref

EndiM:
Addresses issue: #430 

Replacing findDOMNode and string ref usage with callback refs so code run in StrictMode doesn't give warnings.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,6 @@
 
 import React from 'react'
 import PropTypes from 'prop-types'
-import ReactDOM from 'react-dom'
 import classname from 'classnames'
 
 /* Decoraters */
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -309,7 +308,7 @@ class ReactTooltip extends React.Component {
     let desiredPlace = e.currentTarget.getAttribute('data-place') || this.props.place || 'top'
     let effect = switchToSolid && 'solid' || this.getEffect(e.currentTarget)
     let offset = e.currentTarget.getAttribute('data-offset') || this.props.offset || {}
-    let result = getPosition(e, e.currentTarget, ReactDOM.findDOMNode(this), desiredPlace, desiredPlace, effect, offset)
+    let result = getPosition(e, e.currentTarget, this.tooltipRef, desiredPlace, desiredPlace, effect, offset)
```
aronhelser: Is tooltipRef ever set?

EndiM: [tooltipRef is set here](https://github.com/wwayne/react-tooltip/blob/d53942a0c5ab88db828a781281cf0b9b94e2986e/src/index.js#L556)

PR #481, opened 2019-03-21
perf: Convert forEach to plain for

katalonne:


aronhelser:
Does this provide a measurable benefit to you? Can you share any profiling you've done to show this is needed?

PR #487, opened 2019-04-10
fix(index.js): Replaced the deprecated `componentWillPeceiveProps`.

jackall3n:
React has deprecated the use of componentWillReceivedProps, the static method
getDerivedStateFromProps should be used.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -133,15 +133,18 @@ class ReactTooltip extends React.Component {
     this.bindWindowEvents(resizeHide) // Bind global event for static method
   }
 
-  componentWillReceiveProps (props) {
-    const { ariaProps } = this.state
-    const newAriaProps = parseAria(props)
-
+  static getDerivedStateFromProps (nextProps, prevState) {
+    const { ariaProps } = prevState
+    const newAriaProps = parseAria(nextProps)
     const isChanged = Object.keys(newAriaProps).some(props => {
       return newAriaProps[props] !== ariaProps[props]
     })
-    if (isChanged) {
-      this.setState({ ariaProps: newAriaProps })
+    if (!isChanged) {
+      return null
+    }
+    return {
+      ...prevState,
+      ariaProps: newAriaProps
     }
   }
 


aronhelser:
Thanks for the update! It's great to keep up with React.

Unfortunately, you have some linting problems - you can run `make lint` to see them, or look at the Travis output above.

SNikon:
@jackall3n My apologies for intruding on your PR but I believe you are replacing state with its own copy of state at every render call.
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops
I may be incorrect here but after you're correctly checked if ariaProps changed, you should either return null if nothing changed or return the new ariaProps only.

Thanks for helping maintain the library, I'm hoping to use this in a new project

jackall3n:
@SNikon thanks, let me check it out :)

jackall3n:
As per the comment by @SNikon, I've updated the lifestyle method to return `null` when there is no state change.

src/index.js
```diff
@@ -133,15 +133,18 @@ class ReactTooltip extends React.Component {
...
   }
 
-  componentWillReceiveProps (props) {
-    const { ariaProps } = this.state
-    const newAriaProps = parseAria(props)
-
+  static getDerivedStateFromProps (nextProps, prevState) {
+    const { ariaProps } = prevState
+    const newAriaProps = parseAria(nextProps)
     const isChanged = Object.keys(newAriaProps).some(props => {
```
tihuan: Not within the scope of this PR, but it seems like this check doesn't cover the case where `newAriaProps` is a subset of the previous `ariaProps`?

E.g.,

```ts
// prevProps
a = { a: 1, b: 2, c: 3 }

// new props
b = { a: 1, b: 2 }

// returns false, but `a` and `b` actually have different props
const isChanged = Object.keys(b).some(prop => b[prop] !== a[prop])
```

So maybe adding a length check would do?

E.g.,

```ts
const isChanged = Object.keys(newAriaProps).some(props => {
  return newAriaProps[props] !== ariaProps[props]
}) && Object.keys(newAriaProps).length === Object.keys(ariaProps)
```

gfx:
Now React 16.9.0 is released, which produces warnings about `componentWillReceivedProps`.

5minpause:
What's holding this from merging and releasing? Where do you need help?

PR #519, opened 2019-10-27
Add support for shadow DOM elements

### LINKED ISSUE #350: Not working on Shadow DOM
I'm trying to use ReactTooltip in a project with ShadowDOM and it's not showing up, then I tried moving it out of ShadowDOM and it worked. So far I don't know the reason why isn't working. I can only think about CSS which is what ShadowDOM changes the default behavior.

This is the lib I'm using to create the ShadowDOM: https://github.com/Wildhoney/ReactShadow

Does someone faced the same issue and know if it's possible to get it working inside a ShadowDOM?

omerdn1:
The `querySelectorAll` tooltips finder does not work within shadow DOM elements since shadow DOM is created separately from the browser DOM and `querySelectorAll` on the browser DOM will not return any nodes within shadow DOMs.

This PR adds support for shadow DOM elements by scanning the document for shadow DOM elements and appending those elements to the `targetArray`.

This PR fixes #350


aronhelser:
Thank you for your contribution! This seems like a useful addition.

The call to `getElementsByTagName('*')` is a bit concerning - are you making a list of every element in the page? Won't that be slow? If it is slow, we should probably guard this code with an option, so you have to turn it on explicitly.

Second, can you reformat your commit message to start with `fix(selector): ` - this is so semantic release will recognize the commit as needing a new release.

omerdn1:
@aronhelser hey!

Currently, the only way to determine which elements in a given document contain a shadow DOM is to iterate over all of the elements in the document and check if a given element has a `shadowRoot` property.

I used `getElementsByTagName('*')` to get all of the elements in the document, `filter` to apply the necessary `shadowRoot` property check, and `forEach` to iterate over the shadow DOM elements and append the selected target to the `targetArray`.

I'm not creating a list of every element in the page.
The use of `getElementsByTagName('*')` is very similar to `querySelectorAll` and is actually faster. You can read more about it here: https://humanwhocodes.com/blog/2010/09/28/why-is-getelementsbytagname-faster-that-queryselectorall/

Going to reformat my commits.
Thanks Aron.

PR #526, opened 2020-01-02
Feature: listen body events

artdevgame:
Fixes conflict in #273 so we can get this feature merged.

Adds `bodyMode` param that changes behaviour of event listener attachment.


aronhelser:
Awesome. Will you reformat your last commit message so it will trigger sematic-release? Like this:

    feat(bodyMode): fix conflict to finalize bodyMode

PR #528, opened 2020-01-04
fix(getPosition): updated getPosition to fix 'maximum update depth'

kukchanka:
Fix for issue: https://github.com/wwayne/react-tooltip/issues/527


aronhelser:
Thanks! Very clear explanation of the issue, and a quick fix - the best combination 👍

Issue #533, opened 2020-01-11
Event listeners are not freed


mjeffrie:
Event listeners are [created](https://github.com/wwayne/react-tooltip/blob/d1d009cdd597a754a3d929a4ab4c20fb0777a7b3/src/index.js#L511) and sometimes [not removed](https://github.com/wwayne/react-tooltip/blob/d1d009cdd597a754a3d929a4ab4c20fb0777a7b3/src/index.js#L515). This leads to memory leaks.

This happens when `isCaptureMode` is true, since [`removeEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) needs the same `useCapture` flag as used in `addEventListener`.

Possible fixes include:
1. calling `removeEventListener` twice with `true` and `false` for `useCapture`.
2. storing the `useCapture` value used by `addEventListener` and using it when calling `removeEventListener`.
3. providing `currentTarget` to determine `isCaptureMode` when calling `removeEventListener`.

PR #534, opened 2020-01-14
fix: Release event listeners

mjeffrie:
Fix for issue: #533

This patch applied on top of `v3.10.1` works fine in the `make dev` test except for the test scrolling test, which appears to work as it does in master. With a slight tweak, this patch can be applied on master.

@aronhelser Can we get a release with this patch applied to `v3.10.0`, or if not, then `v3.10.1`?

Details on the test scrolling test: The left example should hide the tooltip when scrolling. This happens on the first time, but moving the mouse slightly makes the tooltip reappear, and it does not hide again when scrolling again. This is also how it behaves without this PR's patch.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -169,7 +169,7 @@ class ReactTooltip extends React.Component {
     this.clearTimer();
 
     this.unbindListener();
-    this.removeScrollListener();
+    this.removeScrollListener(this.state.currentTarget);
     this.unbindWindowEvents();
   }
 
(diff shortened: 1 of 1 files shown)


Rogger794:
@aronhelser what do you think about this PR?

aronhelser:
I would use `isCapture` in parallel with the add call just above. The docs on add/removeEventListener make it clear the last boolean needs to match.

Rogger794:
So this PR should be rejected

aronhelser:
I would modify it. It also needs to have conflicts resolved before it can
be applied to the current master - but the submitter was requesting a patch
release based on 3.10? I'm unable to do that - are you willing?

Rogger794:
Yeah, I would check it

mjeffrie:
Updated the PR to determine `isCapture` as `addEventListener` does using the current target as stored in state.

We can take this patch on master and aren't looking for a patch on 3.10 at this point.

Issue #535, opened 2020-01-15
in bodyMode search parents of current target for tooltips


idc-developer:
Switching to bodyMode left our tooltips non functional, since in svg e.g. a textbox often is composed of a <g><rect/><text/></g>, but the tooltip should popup everywhere inside the g, e.g. 
`<g data-tip="my tip"><rect/><text/></g>`
But body mode will currently search the data-tip in rect if you click on the box with no text or <text> if you click on text.

bodyListener should search from e.target the parent chain for data-tips.

PR #530, opened 2020-01-05
Using rollup for building the project, and modifying example

Rogger794:
Hi @wwayne,
I refactored how to build the project by using rollup. So first run the script yarn build. If you would like to try the example, enter into example folder and run yarn to install the dependencies; I add a link to the actual package and the run yarn start. I would like to be a maintainer. I already use this package into one of my packages. https://www.npmjs.com/settings/rogger.794/packages.
I opened for feedbacks.

PR #536, opened 2020-01-15
feat(bodyMode-parentchain) #535 search parent chain for data-tip

idc-developer:
walks the parent chain of the tooltip target to find a data-tip attribute.
It does not check if the mouse position is inside the parents bounding rect, so this only works if children are completely inside the client area of their parents.


aronhelser:
Thanks for your PR! Would you take a look at https://github.com/wwayne/react-tooltip/pull/530 and let me know how much it overlaps, and which might be preferable, for the build process?

idc-developer:
Hi Aron,

thanks for your input, I think I set up the pull requests wrong, it should only include the first commit #dba14c4.
Can you cherry pick this or do I have to resubmit a proper pull request ?

For the conversion from sass to node-sass I think the rollup setup goes further, so prefer #530 and reject mine.


Marco


Von: Aron Helser <notifications@github.com>
Gesendet: Donnerstag, 30. Januar 2020 15:19
An: wwayne/react-tooltip <react-tooltip@noreply.github.com>
Cc: METUS Dev <metus-dev@id-consult.com>; Author <author@noreply.github.com>
Betreff: Re: [wwayne/react-tooltip] feat(bodyMode-parentchain) #535 search parent chain for data-tip (#536)


Thanks for your PR! Would you take a look at #530<https://github.com/wwayne/react-tooltip/pull/530> and let me know how much it overlaps, and which might be preferable, for the build process?

idc-developer:
I will reissue on latest

PR #550, opened 2020-02-07
BREAKING CHANGE: Updating readme for demo

Rogger794:
Using Code sandbox for demo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,10 @@
 [download-image]: https://img.shields.io/npm/dm/react-tooltip.svg?style=flat-square
 [download-url]: https://npmjs.org/package/react-tooltip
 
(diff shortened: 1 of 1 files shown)


markwoon:
What exactly is the breaking change?

Rogger794:
Hi @markwoon, it was about the way it builds. It wasnt updated before.

markwoon:
So there's no actual change in usage?  Just want to confirm.

Rogger794:
Now it is imported as default as the example shows.

wmertens:
Ah ok - could you perhaps explain how it was importing before so people
using the old version know how to change it?

Thanks :)

Wout.

PR #559, opened 2020-03-04
Aphrodite JSS deprecation

kebabmaster:
Hi @Rogger794,

I have replaced the **aphrodite-jss* implementation with a custom styling solution that does not require any external dependencies.

Tested locally and with the **gh-pages** branch, works correctly.

The unit tests have been updated accordingly.

No changes in workflows or documentation.

Thank you for your time and have a wonderful day.


aronhelser:
You're close on the commit message format, but not quite 😄 
`fix(aphrodite_jss_deprecation): aphrodite_jss replaced ...`

kebabmaster:
> You're close on the commit message format, but not quite 
> `fix(aphrodite_jss_deprecation): aphrodite_jss replaced ...`

Argh, supposed to be the easiest part! Fixed.

src/utils/uuid.js
```diff
@@ -0,0 +1,17 @@
+/**
+* By w3resource, 2020, https://www.w3resource.com/javascript-exercises/javascript-math-exercise-23.php
+*/
+
+export function generateUUID(){
+    var dt = new Date().getTime();
```
Rogger794: this could be replace by a let variable

src/utils/uuid.js
```diff
@@ -0,0 +1,17 @@
...
+
+    var uuid = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, (c) => {
+        var r = (dt + Math.random()*16)%16 | 0;
```
Rogger794: r should be const

src/utils/uuid.js
```diff
@@ -0,0 +1,17 @@
+/**
+* By w3resource, 2020, https://www.w3resource.com/javascript-exercises/javascript-math-exercise-23.php
+*/
+
+export function generateUUID(){
+    var dt = new Date().getTime();
+
+    var uuid = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, (c) => {
```
Rogger794: should be let variable

src/utils/uuid.js
kebabmaster: Addressed

src/utils/uuid.js
kebabmaster: Addressed

src/utils/uuid.js
kebabmaster: Addressed

kebabmaster:
Additionally, added the **no-var** **eslint** rule that prevents the further contributors from using **var** instead of **const** or **let**.

PR #566, opened 2020-03-23
fix(uuid): Use uuid package for unique class names

### LINKED ISSUE #562: Generated UUID in 4.0.5 breaks snapshot testing
This commit https://github.com/wwayne/react-tooltip/commit/78e3c2be784f2b96d8bb888b81785b2b38179dae added `generateUUID`, which generates a random UUID every time a tooltip is rendered, which changes Jest snapshots every time they're run.

That commit was done in https://github.com/wwayne/react-tooltip/pull/559, which says "No changes in workflows or documentation", which, while true, apparently didn't take into account changes in the DOM and testing. I would recommend a solution without the `uuid` state, if possible, but if not, some override that would disable the UUID in testing

CrOrc:
Fixes #562


src/utils/uuid.js
```diff
@@ -1,17 +1,5 @@
...
 export function generateUUID(){
-    let dt = new Date().getTime();
-
-    let uuid = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, (c) => {
-        const r = (dt + Math.random()*16)%16 | 0;
-        dt = Math.floor(dt/16);
-        return (c === "x" ? r :(r&0x3|0x8)).toString(16);
-    });
-
-    uuid = "t" + uuid.substring(1, uuid.length); // CSS does not work correctly with classes starting with a numeric character
```
aronhelser: Please leave the comment.

Rogger794: Hi @CrOrc, @aronhelser, could you please check the refactoring branch, I was working on this change by using root variables instead of generating this classes.

imjoshdean-tc:
Any update on this? Would like to merge in v4.1.3 into one of my repositories but our snapshot tests are broken.

aronhelser:
I'm fine with the change, as long as uuid is a small package. Up to you @Rogger794

imjoshdean-tc:
I'd hate to be "that guy" but there should probably be either a snapshot test or some sort of unit test as well, since the whole point of this PR is to fix the issue that this breaks snapshot testing.

Rogger794:
Hi @imjoshdean-tc, could you try the new release?

imjoshdean-tc:
I'll spend some time with it in the next day or so. Is the long term plan to keep using randomly generated strings that can't be overridden? Might not be for me to say, but requiring devs to mock an internal package to the library for testing doesn't feel like a good long-term solution, especially undocumented.

PR #569, opened 2020-03-31
fix(src/decorators/staticMethods.js): fixing another IE event bug

Jarch09:
fixing currentTarget for static method global show (in IE)

After making a partial fix last night for IE edge, this popped up in my error logs...
```
Unable to get property 'target' of undefined or null reference

./node_modules/react-tooltip/dist/index.es.js in t.prototype.globalShow at line 1385:1
  target.prototype.globalShow = function (event) {
    if (this.mount) {
      // Create a fake event, specific show will limit the type to `solid`
      // only `float` type cares e.clientX e.clientY
      var e = {
        currentTarget: event.detail.target
      };
      this.showTooltip(e, true);
    }
  };
```

So this PR has a simple fix to prevent the error.


Jarch09:
Hey guys - can we please get this merged?

Issue #570, opened 2020-04-01
How to set data-event nothing?


seiyab:
How to set data-event nothing?

I want a tooltip that is shown at initial view, and disappears when `data-event-off` is fired.
So I tried to use `ReactTooltip.show` to show, and make no event show the tooltip.

I want to know a way to disable `data-event`, or other implementation to achieve the purpose.

seiyab:
I've already tried `data-event={ null }` and `data-event=""`.
Tooltips were triggered by default event because they were considered as `data-event` is not set .

ReptoxX:
+1 need to find a solution

gleandroj:
Just type anything different of data-event={ null } and data-event="", 
I tried data-event="custom" and works fine!

gleandroj:
+1 need to find a solution

Arshad561:
+1 need some way to disable data-event

PR #573, opened 2020-04-07
[ WIP ] option to disable data-event

seiyab:
Suggestion to solve #570 
`data-event=""` disables trigger.

I'll fix following TODOs when maintainers accept this interface.

TODO:
- body mode (I don't know what is this).
- data-event-off
- documentations
- maybe example


seiyab:
I found `data-event=""` breaks compatibility if someone uses `""` to declare default event.
`data-event="none"` might be better.
I'd like to hear opinions.

PR #582, opened 2020-04-12
Mark prop-types and uuid as external to avoid bundling them

znarf:
prop-types and uuid are currently marked as "dependencies" but are also bundled by rollup.

This PR proposes to configure them as "external" to avoid bundling them.


znarf:
Thank you @roggervalf !

PR #583, opened 2020-04-17
fix: made it possible to pass uuid instead of generating one internally

### LINKED ISSUE #580: Using UUID in class names breaks server-side rendering
The commit https://github.com/wwayne/react-tooltip/commit/78e3c2be784f2b96d8bb888b81785b2b38179dae breaks the server-side rendering of host code, if the host code uses it, because different UUIDs are generated for the same tooltip at the server-side and client-side, and in result the server-generated markup does not match the one calculated client-side.

At the first glance, I don't follow why was it necessary to introduce UUIDs in class names, it looks to me the static class name `__react_component_tooltip` worked perfectly, and there was no need to change that as a part of `aphrodite-jss` deprecation. If you believe UUIDs are really necessary here, they should be exposed as an optional property of tooltip component, so that a user relying on server-side rendering can generate the value himself, then pass and use the same value at the client side.

RamonGebben:
When passing the `uuid` prop ReactTooltip will not generate one itself so you can have a predictable
ID when using SSR strategies"

fix #580

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -86,7 +86,7 @@ class ReactTooltip extends React.Component {
     super(props);
 
     this.state = {
-      uuid: generateUUID(),
+      uuid: props.uuid || generateUUID(),
       place: props.place || "top", // Direction of tooltip
       desiredPlace: props.place || "top",
       type: "dark", // Color theme of tooltip


aronhelser:
Looks like a reasonable solution to me. Can you add it to the docs?

aronhelser:
Also, your commit message must be formatted for sematic-release to automatically trigger:

fix(uuid): make it possible....

PR #613, opened 2020-06-18
#535 search parents of click target in body mode

### LINKED ISSUE #535: in bodyMode search parents of current target for tooltips

idc-developer:
This will fix #535, reissued pull request after build sytem change to rollup

### DIFF
--- a/src/decorators/bodyMode.js
+++ b/src/decorators/bodyMode.js
@@ -19,20 +19,20 @@ const bodyListener = function(callback, options, e) {
   const { respectEffect = false, customEvent = false } = options;
   const { id } = this.props;
 
-  let tip = null
-  let forId
-  let target = e.target
-  let lastTarget
+  let tip = null;
+  let forId;
+  let target = e.target;
+  let lastTarget;
   // walk up parent chain until tip is found
   // there is no match if parent visible area is matched by mouse position, so some corner cases might not work as expected
   while (tip === null && target !== null) {
-    lastTarget = target
-    tip = target.getAttribute('data-tip') || null
-    forId = target.getAttribute('data-for') || null
-    target = target.parentElement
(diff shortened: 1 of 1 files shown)


knubie:
@idc-developer Are you using this branch in production? Do you have the dist hosted somewhere like on github? I'd like to use your branch in production if possible.

idc-developer:
Yes we do. We have forked it and committed the dist into our fork master: https://github.com/idc-developer/react-tooltip
Marco

aa333:
@idc-developer thank you for your work. Would it be possible to add the `data-tip-disable` short return to your walker loop? I'd like to be able to override/disable tooltips on some child elements, i.e. if you meet `data-tip-disable`, you don't need to go up.

Although it's debatable since someone might expect that when you disable tooltip on child, tooltip on parent should take its place.

idc-developer:
> 
> 
> @idc-developer thank you for your work. Would it be possible to add the `data-tip-disable` short return to your walker loop? I'd like to be able to override/disable tooltips on some child elements, i.e. if you meet `data-tip-disable`, you don't need to go up.
> 
> Although it's debatable since someone might expect that when you disable tooltip on child, tooltip on parent should take its place.

The name seems not well fitting. Maybe you should name it like browser events "data-tip-stop-propagation". Currently I have no time to integrate this, but you could fork and do it yourself and create a pull request here.

PR #622, opened 2020-08-03
Check if there is internal ref to prevent calling getBoundingClientRect on undefined.

Baael:
Library is trying to call `getBoundingClientRect` on undefined. Can be reproduces with use of `event="mousedown mouseenter"`.

May be related to https://github.com/wwayne/react-tooltip/issues/531 and few other issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "react-tooltip",
-  "version": "4.2.5",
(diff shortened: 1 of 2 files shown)


landonalder:
Hey @Baael, I'm having the issue this is meant to fix. Would love if your fix got merged. Just curious why this was closed? Thank you for investigating a fix!

PR #629, opened 2020-08-27
fix(getposition): properly determine parents with will-change: transform

KhaledSakr:
Parent elements with property will-change: transform change the containing block, we have to account
for that when calculating the tooltip position. Reference:
https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_Block


aronhelser:
Great fix, thanks!

Issue #631, opened 2020-09-14
Build: Unable to run build via Windows cmd shell. error with cpy


jafin:
Attempting to run the build on Windows cmd shell results in an error for the `copy:types` step.

```shell
d:\react-tooltip>npx cpy './react-tooltip.d.ts' './dist' --parents --dot
Cannot copy `'./react-tooltip.d.ts'`: the file doesn't exist

d:\react-tooltip>npx cpy ./react-tooltip.d.ts ./dist --parents --dot
d:\react-tooltip>
```
If the single quotes are removed then the error is avoided. This should be cross platform safe.

PR #632, opened 2020-09-14
fix(build): enable build via Windows shell

jafin:
Windows shell does not let cpy work with single quotes.  Removed quotes from copy:types fixes(#631)


roggervalf (approved):
Thank you @jafin for this fix!

Issue #469, opened 2019-02-13
My tooltip does not appear


FruitAndAShape:
I have...

```
import ReactTooltip from 'react-tooltip'

<ReactTooltip effect="solid" />

<th data-tip={text} ... />

```
Just like the instructions say. If I uncheck a bunch of CSS rules then I can see the tool tip sitting empty on the screen.

But it doesn't appear when I roll over the <th>. What reason could there be for this?

aronhelser:
If your item is not a child of the ReactToolTip, you need to use the `data-for='happyFace'` property on your item, and the `id='happyFace'` property on the ReactToolTip, so they link up. It's shown in the first example. I'll take a look at the code at the top and see why the first example code shown is not correct.

FruitAndAShape:
That didn't fix it unfortunately.

FruitAndAShape:
It works if I add a tooltip to another page element. Why doesn't it like my generated table header cells?

aronhelser:
It's because your generated table cells aren't present when the Tooltip is
created. You need to call `.rebuild()` after they are created so the
tooltip can discover them.

FruitAndAShape:
Where is the optimum place to do this? After every render?

FruitAndAShape:
I did this...

```
componentDidUpdate() {
   ReactTooltip.rebuild();
}
```

PR #633, opened 2020-09-15
docs(readme): add tip on rebuild for dynamic content relates to #469

jafin:
* Add a note about dynamic rebinding of tooltips to capture lost event listeners as discussed in #469
* fix markdown lint violations.


aronhelser:
Awesome, thanks!

PR #666, opened 2021-01-14
fix(style injection): change style injection default root

### LINKED ISSUE #665: Styles not injecting properly to shadow DOM
Hi,

I've encountered a problem with styles not injecting properly, while using shadow DOM. They don't get applied :(
This was supposed to be fixed in `4.2.11`, but it seems to inject styles into `head` like before. I believe it should inject them into shadow root host.

Thanks in advance!

janeksmielowski:
Change style injection root to evaluated parent node, that should equal to nearest shadow root host,
otherwise use head

fix #665

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -185,7 +185,7 @@ class ReactTooltip extends React.Component {
         parentNode = parentNode.parentNode;
       }
       const head = parentNode.querySelector('head');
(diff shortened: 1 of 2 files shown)


roggervalf (approved):
LGTM, thank you for your contribution

Issue #650, opened 2020-11-23
4.2.11 no longer adds global __react_component_tooltip style


patricekaufmann:
The project I use this on has 2 entry points and it works for one entry point while it does not work for the other one. It used to work for both. The global style tag is missing, containing 

```border-radius: 3px;
    display: inline-block;
    font-size: 13px;
    left: -999em;
    opacity: 0;
    padding: 8px 21px;
    position: fixed;
    pointer-events: none;
    transition: opacity 0.3s ease-out;
    top: -999em;
    visibility: hidden;
    z-index: 999;```

It happens only in 4.2.11, still working in 4.2.10.

gajus:
If anyone runs into this while using next.js, just adding https://github.com/wwayne/react-tooltip/blob/master/src/index.scss to your global styles fixes the issue.

sumthief:
Hey

I think my previous [PR](https://github.com/wwayne/react-tooltip/pull/636) could broke this functionality.

I tried to [fix it](https://github.com/sumthief/react-tooltip/commit/57fc1b3b8652cc0a6acc303f16af27d880106624) (looks like condition for `host` is unnecessary)

Checked on different apps:
* Clean CRA app
* Clean CRA app with tooltip rendered in Shadow DOM
* App inside iframe
* App inside Custom Element
* App inside Custom Element with Shadow DOM

Please could someone check [this fix](https://github.com/sumthief/react-tooltip/commit/57fc1b3b8652cc0a6acc303f16af27d880106624) and give feedback about it?

gjvoosten:
@sumthief That doesn't appear to fix it for us: still shows a black line at the top of the content block (where the tooltip container is), shows the tooltip at the top of the content block instead of where it should be, and the tooltip also no longer hides itself. None of this is the case with 4.2.10.

sumthief:
Okay, It was blind try to fix it and it doesn't work 🙁 

Can someone provide minimal repository for reproducing this issue?

BoresXP:
This happened for me too.
Let's see why it happens. injectStyles should do the trick but it doesn't because [domRoots is empty](https://github.com/wwayne/react-tooltip/blob/master/src/index.js#L191).
This happens because getTargetArray returns empty array :) [here](https://github.com/wwayne/react-tooltip/blob/master/src/index.js#L180).
[...]

PR #668, opened 2021-01-15
fix(styles): change style injection way

sumthief:
As continue of [previous PR](https://github.com/wwayne/react-tooltip/issues/665) and fix for #650.

Previous way doesn't work correct for tooltips without related target elements :slightly_frowning_face:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -185,9 +185,19 @@ class ReactTooltip extends React.Component {
     while (parentNode.parentNode) {
       parentNode = parentNode.parentNode;
     }
-    const head = parentNode.querySelector('head');
 
-    const domRoot = parentNode || head;
+    let domRoot;
+
+    switch (parentNode.constructor.name) {
+      case 'HTMLDocument':
+        domRoot = parentNode.head;
+        break;
+      case 'ShadowRoot':
+      default:
+        domRoot = parentNode;
+        break;
+    }
+
     // Prevent styles duplication.
     if (!domRoot.querySelector('style[data-react-tooltip]')) {
       const style = document.createElement('style');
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -176,29 +176,25 @@ class ReactTooltip extends React.Component {
...
+    while (parentNode.parentNode) {
+      parentNode = parentNode.parentNode;
+    }
+    const head = parentNode.querySelector('head');
+
+    const domRoot = parentNode || head;
+    // Prevent styles duplication.
+    if (!domRoot.querySelector('style[data-react-tooltip]')) {
       const style = document.createElement('style');
       style.textContent = baseCss;
```
louisgrasset, line 194: What about removing spaces & breaklines from output ?
`style.textContent = baseCss.replace(/(\r\n|\n|\r)/gm, '')).replace(/\s+/g, ' ')`

louisgrasset (changes requested):

src/index.js
sumthief: Hey @louisgrasset

IMO css minify should be done in other PR as this PR only changes way of styles injection, neither styles content or form

Also I think css minify should be done via rollup or other tooling but not manually

sumthief:
hey @roggervalf

Looks like we need your attention one more time :slightly_smiling_face:

roggervalf:
> hey @roggervalf
> 
> Looks like we need your attention one more time slightly_smiling_face

hi @sumthief, sorry for the delay, I tried your changes in the current example and it is broken, could you please build your changes and then go to the example example directory:
```sh
yarn build
cd example
yarn
yarn start
```
To check it please

sumthief:
Sorry. Looks like I've tested it wrong way. Now it should be fixed.

PR #670, opened 2021-01-27
chore(dependency): upgrade uuid package to the recent version

Alxblsk:
This PR is upgrading `uuid` to the recent version `8.3.2`.

According to the doc, the only incompatible change is the absence of Deep Requires and Default Exports; none of there features are not used though.

Reference: https://www.npmjs.com/package/uuid


Alxblsk:
@wwayne Can this PR appear as part of the next build?

PR #687, opened 2021-03-24
fix: use delayUpdate in updateTooltip() if tooltip is already shown

### LINKED ISSUE #683: Regression: mouse over tooltip broken by #676
In our application we want to be able to move the mouse over the tooltip to select and copy its content.

With 4.2.15 this is no longer possible because the tooltip instantly moves to follow the mouse, like an unreachable carrot at the end of a stick.

In our application the tooltip is defined within a large component, but the tooltip content depends on the mouse position being over items inside the component that are determined by the application but unknown to react. So using effect='solid' is not an option since this will position the tooltip relative to the large component, not to the mouse position.

PatrickTasse:
fix #683

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -551,7 +551,7 @@ class ReactTooltip extends React.Component {
    * When mouse hover, update tool tip
    */
   updateTooltip(e) {
-    const { delayShow, disable } = this.state;
+    const { delayShow, delayUpdate, disable } = this.state;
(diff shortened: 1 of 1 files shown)


danielbarion:
Hi @PatrickTasse, can you update the PR with the Master and confirm if that one is still needed?

Thanks!

danielbarion:
Closing due to V5

PR #689, opened 2021-03-26
fix: update position if content changed while tooltip is shown

### LINKED ISSUE #638: Wrong tooltip positioning after content text change on click
Hey, I have one small problem with tooltip positioning in case tooltip content has changed while tooltip is displayed. 

It's working if instead of passing function to getContent prop, I pass array. But then  this function is being called every second.

```
function renderContent() {
    return props.tooltipText;
}

return <ReactTooltip
    ...
    // Tooltip position is not updated after text change
    getContent=(renderContent)
    // Tooltip position is updated but its being called every second
    getContent={[renderContent, 1000]}
/>
```

You can try it live:
https://codesandbox.io/s/react-tooltip-example-3116-forked-ydnr7

Any idea how to make it work without calling renderContent function in some interval?

image

PatrickTasse:
fix #638

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -135,6 +135,8 @@ class ReactTooltip extends React.Component {
     this.delayHideLoop = null;
     this.delayReshow = null;
     this.intervalUpdateContent = null;
+    this.prevContent = null;
+    this.content = null;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -149,6 +151,18 @@ class ReactTooltip extends React.Component {
     this.injectStyles(); // Inject styles for each DOM root having tooltip.
   }
 
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (
+      prevState.show &&
+      this.state.show &&
+      this.content &&
+      this.content !== this.prevContent
```
PatrickTasse, line 159: Note to reviewers: I wanted to use a state variable instead of new instance variables so this could be this.state.content !== prevState.content instead, but I got errors and saw recommendations not to call setState() from within render(). If there's a better way to achieve this let me know.

danielbarion:
Closing due to V5

PR #695, opened 2021-04-21
fix(src/index.js): add accessibility support for tabbing

jshaw22:
Add accessibility support for displaying tooltips when tabbing over an element when it receives focus.

More reading:
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
"Custom tooltips, sub-menus, and other nonmodal popups that display on hover and focus are examples of additional content covered by this criterion."

https://www.digitala11y.com/tooltip-role/
"The tooltip typically becomes visible in response to a mouse hover, or after the owning element receives keyboard focus. In each of these cases, authors SHOULD display the tooltip after a short delay."

I added a tabIndex to the example (not in this commit) to show a working example of tabbing and displaying tooltips:
tab react tip

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -136,7 +136,7 @@ class ReactTooltip extends React.Component {
    * For unify the bind and unbind listener
    */
   bind(methodArray) {
-    methodArray.forEach(method => {
+    methodArray.forEach((method) => {
       this[method] = this[method].bind(this);
     });
   }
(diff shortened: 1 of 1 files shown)


roggervalf:
Thank you @jshaw22!

src/index.js
```diff
@@ -296,6 +297,7 @@ class ReactTooltip extends React.Component {
           );
         }
         target.addEventListener('mouseleave', this.hideTooltip, isCaptureMode);
+        target.addEventListener('blur', this.showTooltip, isCaptureMode);
```
muhammadtarek, line 300: Shouldn't on blur triggers `hideTooltip` instead? I believe this caused tooltips to be always displayed when we updated to the latest release

jshaw22: Good catch, my apologies, let me fix this

PR #703, opened 2021-05-20
fix: add aria hidden attribute to style tag

### LINKED ISSUE #691: Accessibility issues with screen reader
I'm having issues with the `react-tooltip` accessibility with screen readers. The library renders the tooltip structure in this way:

image

That causes the screen reader to read all the style tag content before getting to the tooltip's text, which is not desired. 
I guess the style tag should either be moved outside the tooltip's markup or contain `aria-hidden="true"` attribute in order to be skipped by screen readers. 
Is there some other solution to this issue? Thank you.

froston:
Add aria hidden attribute to style tag to avoid screen readers to read its content

fix #691

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -780,7 +780,7 @@ class ReactTooltip extends React.Component {
       .join(' ');
 
     if (html) {
-      const htmlContent = `${content}\n<style>${style}</style>`;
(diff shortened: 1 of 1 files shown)


roggervalf (approved):

roggervalf:
Thank you @froston

PR #706, opened 2021-05-25
fix(compability): add polyfill and change styles

### LINKED ISSUE #677: Most recent version 4.2.14 breaks IE11
Likely from this: https://github.com/wwayne/react-tooltip/pull/668

Causes a `Internet Explorer throws SCRIPT5022: HierarchyRequestError` in IE11

Method of injection is not supported

243083df:
fix #677


package.json
```diff
@@ -132,6 +132,7 @@
     "rollup-plugin-sass": "^1.2.2",
     "rollup-plugin-stylelint": "1.0.0",
     "rollup-plugin-url": "^3.0.1",
+    "sass": "^1.34.0",
```
roggervalf, line 135: just curious, why do we need to change node-sass with sass package here?

243083df: Because old node-sass not compile correctly on windows 10 D: (and node-sass depreacted since 26 October 2020)
But I compared, compiled css not changed.

roggervalf: Thanks @243083df, lgtm

roggervalf (approved):

PR #720, opened 2021-09-21
fix: performance issue caused by excessive use of clearTimeout/Interval

astei:
When a large number of tooltips are mounted, there is an excessive amount of unnecessary `clearTimeout()` and `clearInterval()` calls being made, which cause very slow re-renders. In our case, this PR alone shaved off much the re-render time of one particularly expensive re-render (850ms to 472ms).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -587,10 +587,13 @@ class ReactTooltip extends React.Component {
       }
     };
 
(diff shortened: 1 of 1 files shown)


aronhelser:
Looks good, thanks!

PR #735, opened 2021-11-25
fix: effect and type not properly applied at first render

### LINKED ISSUE #731: Thema/Type is not applied at first render, default dark is set even if the type prop is light
I have the tooltip as this:

```
<ReactTooltip id="downloadTooltip" place="top" title={downloadTooltipTitle} type="light">
...
</ReactTooltip>
```

and first render has it with dark type and it is also not working, if i do some other action and get back to it it has the hardcoded light theme/type and working.

KazuAlex:
Fix incorrect color scheme when use type != "dark" at first render
Same with effect different from "float"

Fix #731

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -93,8 +93,8 @@ class ReactTooltip extends React.Component {
       uuid: props.uuid || generateUUID(),
       place: props.place || 'top', // Direction of tooltip
       desiredPlace: props.place || 'top',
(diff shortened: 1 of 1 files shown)


whiteadi (approved):

KazuAlex:
Hi @aronhelser @roggervalf 
Could you take a look please ? :)

whiteadi (approved):

Issue #329, opened 2017-08-28
Does data-tip support jsx?


bluetidepro:
Am I missing a setting, or does `data-tip` not support JSX? I've tried the following, wth `html` set as true, and not:

```
<div>
  <div data-tip={<Text>Published!</Text>} data-for='tooltip'><Icon iconName="user" /</div>
  <ReactTooltip html id='tooltip' />
</div>
```

&

```
<div>
  <div data-tip={<Text>Published!</Text>} data-for='tooltip'><Icon iconName="user" /</div>
  <ReactTooltip id='tooltip' />
</div>
```

And when I hover both, I get `[ object Object ]`.

And to clarify, I know if I do the following, it does work:

```
<ReactTooltip id='tooltip'>
  <Text>Published!</Text>
</ReactTooltip>
```

But because of the issue with using tooltip's in a modal ([Troubleshooting  - Using tooltip within the modal](https://github.com/wwayne/react-tooltip#1-using-tooltip-within-the-modal-eg-react-modal)), I was hoping to use the `data-tip` to send the jsx to the tooltip.

Let me know if that makes sense, if you need any other info, or if there is a way to send the jsx through the `data-tip`. Thanks so much! :)

dankonino:
Haha was there, colleague pointed me to the tutorial website again. It's the first example ->
you can use it this style 
`<li data-tip data-for="meeting" >
            <span>
              <ReactTooltip
                className={css.tooltip}
                place="top"
                type="light"
                effect="float"
                id="meeting"
                delayHide={2000}>`  I am actually solving that right now :)

bluetidepro:
@dankonino I don't think what you are suggesting will work. As pointed out in my 2nd part, it does work if you add the JSX in the `ReactTooltip`, but I'm trying to put the JSX in the element that has the tip. In the `data-tip`. That, I don't think, works.

Hoping I can get an answer or not, may not be able to use this library if that's the case. Any chance you could chime in, @wwayne? Or maybe @huumanoid?

wwayne:
@bluetidepro `data-tip` it actually makes use of the dangerouslySetInnerHTML https://github.com/wwayne/react-tooltip/blob/master/src/index.js#L455
So, yes, JSX doesn't work for `data-tip`

bluetidepro:
Ah, okay. Thanks for the reply @wwayne!

srgraham:
[...]

PR #736, opened 2021-12-06
Added note and example about JSX to Readme

stephanbogner:
#### Proposed change:
Adding a note and example regarding using `data-tip` with JSX to Readme.

#### Reasoning
It took me quite some trial and error until I finally found issue #329.
Also JSX gets more and more common and it seemed there was a fair bit of "traffic" in that issue so it's probably worth adding to the Readme.


aronhelser:
Thanks!

PR #742, opened 2022-02-09
fix(aftershow): call afterShow only after state has fully updated

LennartSpitzner:
`afterShow` was called in the callback of a `setState`, but in the same callback `updatePosition` was also called which did another `setState`. Due to the asynchronous nature of setState that means that `afterShow` did not actually see the actual updated state. This fixes it by forwarding/moving the callback into the innermost `setState` callback. It is a bit annoying/nested but works. Suggestions welcome. Moving to hooks might be the appropriate cleanup, but that would be a somewhat larger task.

Thanks!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -581,10 +581,11 @@ class ReactTooltip extends React.Component {
             show: true
           },
           () => {
-            this.updatePosition();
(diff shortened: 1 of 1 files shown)


aronhelser:
LGTM, thanks!

PR #743, opened 2022-02-09
feat(getposition): support desired-place-list (by priority)

LennartSpitzner:
This adds the feature that `desiredPlace` is not a single place, but (can be) a comma-separated list of places, where the first "inside" place (i.e. that does not clip the window/viewport) is chosen. This also refactors `getPosition` code a bit. The important realisation is that all the current code does is to choose the first "inside" position from the list `[desiredPlace, place, "top", "bottom", "left", "right"]`. From that perspective, allowing `desiredPlace` to be a list is trivial.

Please let me know if there is interest in merging this. I guess the docs would need updating.


aronhelser:
Yes, I do think this is interesting, if you can update the docs, I'll merge it.

LennartSpitzner:
I have updated the docs. Also changed the description of the `offset` option to make it more clear what fields are expected.

PR #745, opened 2022-03-10
feat: add borderClass prop to allow a custom border class

dannyharding10:
Allow overriding the default border class name "border" with a custom class name. Currently, when using a border it adds the class name "border", which is pretty generic and can be overridden by other styles (notably, bootstrap uses the class "border", overriding any `borderColor` provided to react-tooltip). This change makes it so that the user can provide a custom class name for the border class.


danielbarion:
Hi @dannyharding10, can you update your PR with the master branch, please?

I'll take a look at that one after it's updated with Master.

Thanks!

danielbarion:
Thank you @dannyharding10!

That improvement will be released together in the next version, we will probably release that soon.

PR #751, opened 2022-03-29
fix: set aria-describedby value wrong when custom id

Xubin-Hiretual:


Xubin-Hiretual:
When I set an id to props, the screen reader didn't work. I found that aria-describedby is the value of uuid, not the id value I passed.

PR #760, opened 2022-04-12
fix: Its instance type 'ReactTooltip' is not a valid JSX element.

DenisGold:
Fix new problem with error 'Its instance type 'ReactTooltip' is not a valid JSX element.' and 'Type 'React.ReactNode' is not assignable to type 'import("/node_modules/@types/react-dom/node_modules/@types/react/index").ReactNode'.'

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,7 +32,7 @@ import 'core-js/modules/es.array.find';
 @getEffect
 @bodyMode
 @trackRemoval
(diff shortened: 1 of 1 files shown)


timzand (approved):

DenisGold:
it's should be for react 18, lower working fine

rkjain119 (approved):
Tested working please merge

PR #806, opened 2022-10-19
feat: Refactored the arrow implementation, added custom corners support

erezcarmel:
- Refactored the arrow implementation to use a rotated square element instead of border to enable the border-radius support.
- Added custom corners support to the tooltip and arrow.
- Updated tests to support new implementation and styling.


danielbarion:
Looks good to me!

Thanks for the PR!

PR #838, opened 2022-12-08
Fix anchor element selector to support usage of `React.useId()`

### LINKED ISSUE #836: [BUG] Unable to use `React.useId()` hook as id for anchor element
**Describe the bug**
When using React's `useId()` hook as the id for the anchor element, I'm faced with the following error:

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#:r1b:' is not a valid selector.
```

This is because this component is using ```document.querySelector(`#${anchorId}`)``` to access the anchor DOM element, and ids generated by `useId()` (such as `:r1b:`) lead to an invalid selector (`'#:r1b:'`).

**Version of Package**
v5.0.0

**To Reproduce**
- Generate an id with `React.useId()`.
- Pass it to `anchorId`.
- The error occurs.

**Additional context**
Since in v5 the component itself is using the `useId()`, there's no reason why we shouldn't be able to use for anchor ids as well.

One of the possible solutions is using the ```document.querySelector(`[id='${anchorId}']`)``` syntax instead, but I believe there are a few others.

gabrieljablonski:
Closes #836.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -106,7 +106,7 @@ const Tooltip = ({
   const debouncedHandleHideTooltip = debounce(handleHideTooltip, 50)
 
   useEffect(() => {
-    const elementReference = document.querySelector(`#${anchorId}`)
+    const elementReference = document.querySelector(`[id='${anchorId}']`)
 
     if (!elementReference) {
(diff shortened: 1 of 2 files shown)


danielbarion:
Thanks for the PR @gabrieljablonski!

danielbarion (approved):

PR #841, opened 2022-12-08
Add `positionStrategy` prop

### LINKED ISSUE #840: [BUG] Clipping when parent container has `position: relative` and `overflow: hidden`
**Describe the bug**
When using the component inside a container with `position: relative` and `overflow: hidden`, the tooltip is clipped.

**Version of Package**
v5.0.0

**To Reproduce**
- Create a container with `position: relative` and `overflow: hidden`.
- Create an element inside it to be used as the anchor.
- Create a `<Tooltip />` inside it.
- The tooltip will be clipped.

**Screenshots**
image

**Additional context**
This is due to the tooltip using `position: absolute`. The obvious workaround is removing the `position: relative` from the parent, but that's not always viable.

gabrieljablonski:
Closes #840.


danielbarion (approved):

gabrieljablonski:
Any idea why this check started failing? Thought it was an issue with my environment, no idea why it suddenly started happening.

Adding and setting up `eslint-import-resolver-typescript` solves it. If that's ok, I can push the changes.

danielbarion:
Yeah, this was weird, I'm checking your locally. I had to stop because of my work, but I'll check as soon as possible to confirm.

Also, thank you so much for the reports and for the PRs!

Quick question: why you removed the `import type` (type) from import? - I'm just asking because I don't know the side-effect of using or not using type when importing decorators.

And I'm already building the types, how can I improve this? The types are working on a study project that I have.

gabrieljablonski:
> why you removed the `import type` (type) from import?

Oh, this is definitely the issue. Wasn't intentional, I'll revert it.
Basically `import type` gets completely erased after transpilation, so it doesn't show up at runtime. The issue with eslint is probably because I was trying to just `import` a `d.ts` file, but not exactly sure if that's the whole reason.

danielbarion:
Oh, this was the issue hahaha

Thanks @gabrieljablonski, merged 😄

PR #845, opened 2022-12-08
Tooltip provider for easily sharing tooltip amongst multiple elements

### LINKED ISSUE #837: [FEAT REQ] Easier way to attach tooltips to dynamically generated elements
In v4, I was able to create a single `<ReactTooltip />` somewhere on my application, and use the `data-for` prop to share that same tooltip across all text elements (for example) in a table. Since it doesn't seem to be any obvious way to do this in v5, I'm left with using wrapper components that generate an id (#836 is quite relevant here) for my anchor element, alongside the `<Tooltip />` element.

I'm not sure if there is another way to do this aside from the `data-*` method, but it would definitely be very helpful, since that was a very convenient feature in v4. Anyway, I'll understand if this is something the package is trying to shy away from completely in v5.

Another alternative that would make things easier, would be to go the `@mui/material` route, and be able to use the tooltip as a wrapper to the anchor element, but I'm not sure how viable that would be for the v5 architecture.

### LINKED ISSUE #847: One tooltip, many anchors
One of the big advantages of `react-tooltip` <= v4 was you could have a single tooltip somewhere in the root of your application and just scatter the `data-tip` attribute around your application and it would "just work". It looks like having a specific `anchorId` -> `id` relationship in v5 enforces a 1:1 relationship. Unless undocumented, can we support or provide non-verbose ways to achieve this? Thanks.

gabrieljablonski:
Closes #837 
Closes #847

Using just `anchorId` without the provider should still work fine.
See [App.tsx](https://github.com/ReactTooltip/react-tooltip/blob/6d647dcbb28898a88f80088d8f196a8cc0f86cda/src/App.tsx) for usage example.

The only major feature missing is using a prop such as `data-tooltip-id` on the anchor element and filtering the event listener and observer subscriptions by the tooltip id. 
The provider also has to keep the `anchorRefs` and `activeAnchor` separate by tooltip id. 

Instead of using `data-tooltip-id`, it would also be possible to change the context data signature  to something such as
```ts
interface TooltipContextData {
  getAnchorRefs: (tooltipId?: string) => Set<React.RefObject<HTMLElement>>
  attach: (...ref: React.RefObject<HTMLElement>[], tooltipId?: string) => void
  detach: (...ref: React.RefObject<HTMLElement>[], tooltipId?: string) => void
  getActiveAnchor: (tooltipId?: string) => React.RefObject<HTMLElement>
  setActiveAnchor: React.Dispatch<React.SetStateAction<React.RefObject<HTMLElement>>>
}
```

`tooltipId` is used to keep track of which tooltip is being accessed.

Without a change like one of these two, the provider only works with one `<Tooltip />` element.

If needed, multiple providers (one `<Tooltip />` inside each) can be separately used as a workaround.


src/components/TooltipController/TooltipController.tsx
```diff
@@ -125,20 +118,25 @@ const TooltipController = ({
...
-    if (!anchorId) {
-      // eslint-disable-next-line @typescript-eslint/no-empty-function
-      return () => {}
+    if (content) {
+      setTooltipContent(content)
+    }
+    if (html) {
+      setTooltipContent(html)
     }
+  }, [content, html])
```
danielbarion, line 127: Much better, thanks!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -28,10 +28,10 @@ const Tooltip = ({
   isOpen,
   setIsOpen,
 }: ITooltip) => {
-  const tooltipRef = createRef()
-  const tooltipArrowRef = createRef()
-  const tooltipShowDelayTimerRef = useRef<ReturnType<typeof setTimeout>>()
-  const tooltipHideDelayTimerRef = useRef<ReturnType<typeof setTimeout>>()
+  const tooltipRef = useRef<HTMLElement>(null)
+  const tooltipArrowRef = useRef<HTMLDivElement>(null)
```
gabrieljablonski, line 32: was `createRef()` being used intentionally? didn't seem like it. `useRef()` is preferred when using functional components

src/components/Tooltip/Tooltip.tsx
```diff
@@ -28,10 +28,10 @@ const Tooltip = ({
...
+  const tooltipShowDelayTimerRef = useRef<NodeJS.Timeout | null>(null)
+  const tooltipHideDelayTimerRef = useRef<NodeJS.Timeout | null>(null)
```
gabrieljablonski, line 34: it is usually recommend to add a default value to `useRef()`

src/components/TooltipController/TooltipControllerTypes.d.ts
```diff
@@ -26,3 +26,18 @@ export interface ITooltipController {
...
+    'data-tooltip-html'?: string
+    'data-tooltip-variant'?: VariantType
+    'data-tooltip-offset'?: number
+    'data-tooltip-wrapper'?: WrapperType
+    'data-tooltip-events'?: EventsType[]
+    'data-tooltip-position-strategy'?: PositionStrategy
+    'data-tooltip-delay-show'?: number
+    'data-tooltip-delay-hide'?: number
+  }
+}
```
gabrieljablonski, line 43: this will be relevant after figuring out how to emit all types to the build `react-tooltip.d.ts` file.

Autocomplete doesn't work too well with attributes with a `-` in them, but this should enable autocomplete for the values, and throw a compile error when using invalid values.

For example, an element such as `<div data-tooltip-place="botom" />` would not compile (`botom` is not a valid value for `place`)

src/components/Tooltip/Tooltip.tsx
danielbarion: yeah, you're right!

can you do the change and test if the behavior still as expected, please?

src/components/TooltipController/TooltipControllerTypes.d.ts
danielbarion: Oh, got it!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -130,58 +149,80 @@ const Tooltip = ({
...
-  }, [show, isOpen, anchorId])
+  }, [])
 
   return (
     <WrapperElement
       id={id}
       role="tooltip"
       className={classNames(styles['tooltip'], styles[variant], className, {
-        [styles['show']]: isOpen || show,
+        [styles['show']]: !calculatingPosition && (isOpen || show),
```
gabrieljablonski, line 215: The current lifecycle of the tooltip with multiple anchors:

1. New active anchor is set
2. Position calculation (in `Tooltip`) and data attributes observer (in `TooltipController`) effets are triggered
3. New data attributes are set
4. Position calculation resolves, and position-related styles are set

Since 3 and 4 happen asynchronously, they resolve at different times (I believe 3 always resolves first), resulting in the tooltip rendering twice. 

`calculatingPosition` is used to hide the tooltip until 4 resolves, which ensures data attributes and the position have been updated.

This feels like a very hacky solution, so we should probably find a better way to make 3 and 4 happen on the same render cycle.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -64,12 +68,19 @@ const Tooltip = ({
     }, delayHide)
   }
 
-  const handleShowTooltip = () => {
+  const handleShowTooltip = (e?: Event) => {
+    if (!e) {
```
danielbarion, line 72: I believe we should use explicit words, like `event` instead of `e`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -64,12 +68,19 @@ const Tooltip = ({
...
+      return
+    }
     if (delayShow) {
       handleShowTooltipDelayed()
     } else {
       handleShow(true)
     }
+    setActiveAnchor((anchor) =>
+      anchor.current === e.target ? anchor : { current: e.target as HTMLElement },
```
danielbarion, line 81: I believe we should use explicit words, like `event` instead of `e`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -64,12 +68,19 @@ const Tooltip = ({
...
+    )
+    setProviderActiveAnchor({ current: e.target as HTMLElement })
```
danielbarion, line 83: I believe we should use explicit words, like `event` instead of `e`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -102,14 +113,22 @@ const Tooltip = ({
...
+    }
 
-    if (!elementReference) {
+    if (!elementRefs.size) {
       // eslint-disable-next-line @typescript-eslint/no-empty-function
       return () => {}
     }
 
-    const enabledEvents: { event: string; listener: () => void }[] = []
+    const enabledEvents: { event: string; listener: (e?: Event) => void }[] = []
```
danielbarion, line 131: I believe we should use explicit words, like `event` instead of `e`

danielbarion (changes requested):
The looks nice to me, but only 2 things:

I believe we should use explicit words, like `event` instead of `e`.

And can you create docs inside `docs` folder related to the new possibility of use of the Tooltip, please?

src/components/TooltipProvider/TooltipWrapper.tsx
```diff
@@ -27,25 +26,23 @@ const TooltipWrapper = ({
...
+      data-tooltip-offset={offset}
+      data-tooltip-wrapper={wrapper}
+      data-tooltip-events={events}
+      data-tooltip-position-strategy={positionStrategy}
+      data-tooltip-delay-show={delayShow}
+      data-tooltip-delay-hide={delayHide}
+    >
+      {children}
+    </span>
+  )
```
gabrieljablonski, line 45: After some testing, I've realized `cloneElement()` doesn't work super well when passing functional components as children to the wrapper.

Using an actual DOM element (`span` seems ok) didn't seem to cause any issues.

danielbarion (approved):
looks good to me!

PR #861, opened 2022-12-15
feat: add v4's `float` effect (follow mouse position)

gabrieljablonski:
### TODO
- Create and handle `float: boolean` prop to add `mousemove` event listener (is there really a need for an `effect: 'solid' | 'float'` prop?)
- Handle offset according to `place` value
- `handleShowTooltip()` should also use virtual element when `float` is `true`


danielbarion:
I believe we really need to give the option to use a fixed tooltip or this one, both situations are for different use cases.

Also, do you know if for float option is there a way to not let the arrow corner go out of the tooltip container like the image?

image

The implementation looks good to me.

gabrieljablonski:
> I believe we really need to give the option to use a fixed tooltip or this one, both situations are for different use cases.

For sure. I just meant instead of v4's `effect` prop to select which mode to use, we could simply add a boolean `float` or `followMouse` to enable this mode.

> Also, do you know if for float option is there a way to not let the arrow corner go out of the tooltip container like the image?

It'll probably not be too hard to add a minimum distance between the tooltip border and the arrow position.

docs/docs/upgrade-guide/changelog-v4-v5.md
```diff
@@ -8,74 +8,76 @@ If you are using V4 and want to use V5, please read this doc.
...
-- [ ] disable - **Deprecated** | in V5 -> state can be controlled or uncontrolled
-- [ ] scrollHide - not implemented yet, if many users need this feature, we will validate if we wrok on this one.
-- [ ] resizeHide - not implemented yet, if many users need this feature, we will validate if we wrok on this one.
-- [x] wrapper - `data-wrapper`
-- [ ] bodyMode - **Deprecated**
-- [ ] clickable - **Deprecated** | Supported by default in V5
-- [ ] disableInternalStyle - **Deprecated** | in V5 -> CSS will be a separate file and can be imported or not
+- [x] `children`
+- [x] `place` - also available on anchor element as `data-tooltip-place`
+- [ ] `type` - use `variant`. also available on anchor element as `data-tooltip-variant`
```
gabrieljablonski, line 46: Having both the empty brackets `[ ]` and `**Deprecated**` seemed redundant.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -151,15 +217,37 @@ const Tooltip = ({
...
+
+    if (float) {
+      if (lastFloatPosition) {
+        /*
+          Without this, changes to `content`, `place`, `offset`, ..., will only
+          trigger a position calculation after a `mousemove` event.
+
+          To see why this matters, comment this line, run `yarn dev` and click the
+          "Hover me!" anchor.
+        */
```
gabrieljablonski, line 244: For future reference!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -151,15 +217,37 @@ const Tooltip = ({
...
+        handleTooltipPosition(lastFloatPosition)
+      }
+      // if `float` is set, override regular positioning
+      return () => null
+    }
```
gabrieljablonski, line 249: For future reference!

danielbarion (approved):

danielbarion (approved):

PR #873, opened 2022-12-21
Minor bug fixes and docs improvements

### LINKED ISSUE #872: [BUG] Tooltips as react children element don't show anything
**Describe the bug**
The tooltip is not displayed when it is created with custom html (React children elements). The [children  example](https://react-tooltip.com/docs/examples/children) is broken too.

**Version of Package**
v5.3.0

**To Reproduce**
```jsx
<anyElement id="element-id" />

<Tooltip anchorId="element-id">
  <p>should work</p>
</Tooltip>
````

➡️ **Workaround** ⬅️

Just add `content="."` and it should works for now.

```jsx
<anyElement id="element-id" />

<Tooltip anchorId="element-id" content=".">
  <p>now work</p>
</Tooltip>
````

**Expected behavior**
Should show tooltip

 - OS: Windows
 - Chrome
 - Latest

### LINKED ISSUE #875: [BUG] Changing the content prop to a falsey value doesn't update the displayed content
**Describe the bug**
If you control the `content` prop with conditional logic, only the initial value is used for the actual display content.

**Version of Package**
v5.3.0

**To Reproduce**
Create a tooltip like so:
```
const [state, setState] = useState(true);

return (
  <>
    <div id="some-id" onClick={() => setState(!state)}>text</div>
    <Tooltip anchorId="some-id" content={state ? 'some content' : ''}/>
  </>
)
```

**Expected behavior**
Changing the content prop should cause the displayed tooltip to change to `''` (no tooltip). Instead, it will stay as "some content" in the above example regardless of the value of `state`

**Screenshots**
None for now.

 - OS: macOS
 - Browser: Brave (Chromium-based)
 - Version 1.46.144

**Additional context**
This is a real blocker for me because if I try to use the `content="."` hack to fix #872 and use `children` for the conditional instead, then if my conditional returns `null`, then a tooltip is shown with that period in it instead of nothing, like I want. So there is no way to show conditional logic that might return `null`, so there is no way to conditionally disable a tooltip that I know of.

### PR DESCRIPTION
**Pull request by gabrieljablonski**
Closes #872 
Closes #875 

After agreeing on these the PR should be ready to review.

### [State/Uncontrolled tooltip state](https://react-tooltip.com/docs/examples/state#uncontrolled-tooltip-state) section

Should we remove this section? It just shows a limitation of the component (which could be fixed, but why?), and the use case in the example can be much more conveniently achieved by using the recently added provider.

### [Click event](https://react-tooltip.com/docs/examples/events)

Since the recent design change (when clicking an open tooltip with `click` event enabled, it no longer closes the tooltip), does it even make sense to use both `hover` and `click` event simultaneously? Maybe we should remove the last section.

--------------

### Further considerations (slightly related to PR)

I propose a breaking API change (maybe initially just deprecate `events` instead of fully removing it) similar to how V4s `effect: 'solid' | 'float'` prop got replaced by just the `float: boolean` prop.

A `click: boolean` prop would replace the `events` prop to activate the "on click" mode.

### CODE REVIEW

**Review by gabrieljablonski — COMMENTED** [review]
- Since `data-tooltip-content` was never intended to render `html` stuff, removed it from relevant sections.

- The "Content" example page is redundant, since it became exactly the "Basic examples" page.

**Comment by danielbarion** [comment]
> * Since `data-tooltip-content` was never intended to render `html` stuff, removed it from relevant sections.
> * The "Content" example page is redundant, since it became exactly the "Basic examples" page.

I wrote the `content` example page because of the prop, not because of the feature (the `content` can be tracked in algolia search engine, so, if someone wants know about this specific prop, this will be possible because of this page)

**Comment by danielbarion** [comment]
Everything else, I totally agree

File: src/components/TooltipController/TooltipController.tsx [hunk 1]
```diff
@@ -110,10 +109,11 @@ const TooltipController = ({
...
   useEffect(() => {
-    if (content) {
-      setTooltipContent(content)
-    }
+    setIsHtmlContent(false)
+    setTooltipContent(content)
     if (html) {
+      // html will take precedence
+      setIsHtmlContent(true)
       setTooltipContent(html)
```
danielbarion, line 117: I have the feeling that we will spend resources (a bit) doing the:

```
setIsHtmlContent(false)
```

and after that:

```
setIsHtmlContent(true)
```

In this case, I prefer not to do the `setIsHtmlContent(false)` if we will do `setIsHtmlContent(true)` two lines below.

Is this necessary in this order? or can we add a validation to do one or other?

gabrieljablonski: The validation would only serve to make the code clearer (which I think is unnecessary, since it's simple enough).

React 18 batches `useState()` updates, so the state is effectively only updated once. [This source](https://blog.bitsrc.io/automatic-batching-in-react-18-what-you-should-know-d50141dc096e#:~:text=React%20v18%20ensures%20that%20state%20updates%20invoked%20from%20any%20location%20will%20be%20batched%20by%20default.) is a bit outdated, but I'm pretty sure that still holds true.

We can come back to this in the future if we ever decide to improve the `content`/`html` props (maybe there shouldn't be a single state variable - `tooltipContent` - that holds the value for the one being actually used).

danielbarion: oh, interesting!

ok

danielbarion (approved):

PR #881, opened 2022-12-30
fix: package main should be cjs format. (fixes #869)

### LINKED ISSUE #869: Jest fails on standard import syntax
When importing Tooltip, Jest throws the following error unless you specifically import the cjs version.
react-tooltip v5.2.0
To reproduce, in a project running Node 18, React 18, and Jest 29, import tooltip:
```
import { Tooltip as ReactTooltip, ITooltip } from 'react-tooltip';
```
Project runs fine, but jest tests fail.

When running jest tests I see the following error:
```
Jest encountered an unexpected token
    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
...node_modules/react-tooltip/dist/react-tooltip.esm.min.js:1
```
So, it's possible it's just my jest configuration that is unable to support this syntax, but the tests run fine with standard imports from other react ui packages. 

Current solution: directly import from the dist directory's .cjs version.
```
import { Tooltip as ReactTooltip, ITooltip } from '../../node_modules/react-tooltip/dist/react-tooltip.cjs';
```

KholdStare:
Closes #869 

- Add module entry for esm in package.json

See https://rollupjs.org/guide/en/#publishing-es-modules:

> To make sure your ES modules are immediately usable by tools that work
> with CommonJS such as Node.js and webpack, you can use Rollup to
> compile to UMD or CommonJS format, and then point to that compiled
> version with the main property in your package.json file. If your
> package.json file also has a module field, ES-module-aware tools like
> Rollup and webpack 2+ will import the ES module version directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "prepare": "husky install",
     "test": "jest"
   },
-  "main": "dist/react-tooltip.esm.min.js",
+  "main": "dist/react-tooltip.cjs.min.js",
+  "module": "dist/react-tooltip.esm.min.js",
   "buildFormats": [
     {
       "file": "dist/react-tooltip.umd.js",


gabrieljablonski:
I'm not the most familiar with rollup, so we'll just take a look to make sure this doesn't break anything, though this is probably fine as is.

Thanks for the contribution.

danielbarion:
hey o/

my fault, I totally forgot about this, I've worked in the past with an internal react components library for the company I worked for. This will probably work as expected.

I and @gabrieljablonski will test some use cases to double-check, thank you so much!

danielbarion:
Hi guys, version `5.3.3-beta.0` was published from this PR, let's test and give feedback please, if it's working as expected 😄

gabrieljablonski (approved):

itsikcircle:
Still have this issue since `"type": "module",` still exists in the `package.json` file

> react-tooltip.cjs.min.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.

danielbarion:
@itsikcircle yeah, this is probably related like: https://github.com/ReactTooltip/react-tooltip/issues/887

Issue #887, opened 2023-01-03
[BUG] Package fails to import on Next 13


brunocangs:
**Describe the bug**
Next 13 tries to import the package as an ES Module as defined by the `type: module` field in package.json but fails, with an error as the `main` file is `.cjs`.

**Version of Package**
v5.3.3

**To Reproduce**
- Create a new NextJS app with `create-next-app`
- Install `react-tooltip`
- Setup the package and add a tooltip to the `pages/index.ts` file.

**Expected behavior**
The package's `type` and `main` should match their type

**Screenshots**
![image](https://user-images.githubusercontent.com/30100875/210382677-9c60ecf8-a9ee-4b66-84cc-c06402d8ce03.png)

**Desktop (please complete the following information if possible or delete this section):**
 - OS: MacOS 13.0.1

danielbarion:
🤦🏻‍♂️ 

This makes sense, can you do a PR, please?

Thanks for the report!

itsikcircle:
I've been facing the same issue, and send a PR here https://github.com/ReactTooltip/react-tooltip/pull/888

danielbarion:
PR merged and released.

Thanks everyone!

https://github.com/ReactTooltip/react-tooltip/releases/tag/v5.3.4

PR #888, opened 2023-01-03
Fix package.json

itsikcircle:
A fix for
- https://github.com/ReactTooltip/react-tooltip/issues/887
- https://github.com/ReactTooltip/react-tooltip/pull/881

> react-tooltip.cjs.min.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,5 +1,5 @@
-import fs from 'fs'
-import rimraf from 'rimraf'
+const fs = require('fs') // eslint-disable-line @typescript-eslint/no-var-requires
+const rimraf = require('rimraf') // eslint-disable-line @typescript-eslint/no-var-requires
 
 const args = process.argv.slice(2)
 const parameters = args.reduce((acc, arg) => {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     }
   ],
   "types": "dist/react-tooltip.d.ts",
-  "type": "module",
   "license": "MIT",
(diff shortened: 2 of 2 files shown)


itsikcircle:
@danielbarion it should fix it.

MehediH (approved):

dave-circle (approved):

yliu-circle (approved):

itsikcircle:
> Thanks, guys, I'll run some checks locally and do a release today with this PR

Thank you!

danielbarion:
This change is breaking the build, I'll take a look and I hope to finish soon
image

itsikcircle:
> This change is breaking the build, I'll take a look and I hope to finish soon image

It's in the `cli.js` file, it's using ES6 import but it's not supported without `type: module` in the `package.json`
A simple solution is to use `require` instead of `import`, but then eslint will complain.

I've updated the PR with eslint-disable as this is a blocker for us and we can't downgrade to v4.

I can help looking into a better solution without eslint-disable later on.

danielbarion (approved):

danielbarion:
I tested locally and it's working on a next.js project as expected, thanks for the PR!

danielbarion:
I tested locally and it's working on a next.js project as expected, thanks for the PR!

Issue #898, opened 2023-01-11
[FEAT REQ] afterShow


Oupsla:
**Is your feature request related to a problem? Please describe.**
Hello 👋  I was using react-tooltip v4 and the afterShow callback to send an event to our analytics system to analyse the usage of tooltip, but now I don't see any alternative to do that with the v5.

**Describe the solution you'd like**
Put back a callback method 

**Describe alternatives you've considered**
Stay in v4 😢 

**Additional context**
Can help if this require some 😄

PR #899, opened 2023-01-11
feat(Tooltip): implement afterShow callback

Oupsla:
fix: https://github.com/ReactTooltip/react-tooltip/issues/898

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -27,6 +27,7 @@ const Tooltip = ({
   clickable = false,
   style: externalStyles,
   position,
+  afterShow,
   // props handled by controller
   isHtmlContent = false,
   content,
@@ -51,6 +52,10 @@ const Tooltip = ({
       setIsOpen(value)
     } else if (isOpen === undefined) {
       setShow(value)
+    }
+
+    if (value && afterShow) {
+      afterShow()
     }
   }
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -165,6 +165,16 @@ function App() {
           />
         </div>
       </div>
+
+      <div style={{ marginTop: '1rem' }}>
+        <button id="buttonAfterShow">Check the dev console</button>
+        <Tooltip
+          place="bottom"
+          anchorId="buttonAfterShow"
+          afterShow={() => console.log('Hello world')}
+          content="Showing tooltip and calling afterShow method"
+        />
+      </div>
     </main>
   )
 }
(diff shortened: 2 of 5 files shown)


gabrieljablonski:
Hey, thanks for the contribution. 

Since we're already here, would you mind doing `afterHide` as well?

Oupsla:
@gabrieljablonski done 😄

gabrieljablonski:
Thanks. At first glance everything looks good. I'll test it out and review the PR later today.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -52,6 +54,14 @@ const Tooltip = ({
...
       setShow(value)
     }
+
+    // Callbacks
+    if (value === true && afterShow) {
+      afterShow()
+    }
+    if (value === false && afterHide) {
+      afterHide()
+    }
```
gabrieljablonski, line 64:
```suggestion
    if (value) {
      afterShow?.()
    } else {
      afterHide?.()
    }
```

gabrieljablonski (changes requested):
There's a slight issue that should probably be solved.

1. When using `afterHide` on a tooltip with `events={['click']}`, clicking anywhere outside the anchor element will trigger a `handleShow(false)` call, which in turn triggers the `afterHide()` call, which I believe is not desirable.

2. Similarly, when using `afterHide` on a tooltip with `delayShow`, hovering away from the tooltip before it is shown will trigger a call to `afterHide()`.

There should be a check if the tooltip was actually being shown before calling `afterHide()`. Doing the same for `afterShow()` should probably also be done, although I don't believe it will make a difference in practice.

Oupsla:
By using useEffect, it seems to work nicely, tell me if I had understand correctly the bugs and if it is fixed. 
Below some videos with my tests

> events={['click']}

https://user-images.githubusercontent.com/15279620/212007258-a5e5ffa0-c621-46c2-8650-f59228e37692.mov

> delayShow

https://user-images.githubusercontent.com/15279620/212007850-860291da-b830-4e60-8968-b343c7478502.mov

gabrieljablonski (approved):

gabrieljablonski:
> By using useEffect, it seems to work nicely, tell me if I have understood correctly the bugs and if it is fixed.

This fixed the problems I mentioned, but it still left the issue of the `useEffect()` being called on the first render. 

There were a few other ways to fix this, but I ended up adding a `wasShowing` ref to ensure the callbacks are only called when `show` actually changes value.

gabrieljablonski:
Everything should be good to go. As soon as @danielbarion reviews this, I'll merge it and do the v5.5.0 release.

danielbarion (approved):

danielbarion:
The code looks good to me.

Thanks for the implementation @Oupsla!

Thanks, @gabrieljablonski, and feel free to merge when you want.

PR #901, opened 2023-01-11
docs: update README.md to include bundlephobia metrics

danielbarion:
https://bundlephobia.com/package/react-tooltip@5.4.0

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # react-tooltip
 
 [![Version](http://img.shields.io/npm/v/react-tooltip.svg)](https://www.npmjs.org/package/react-tooltip)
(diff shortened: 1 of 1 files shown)


danielbarion:
@gabrieljablonski I added the link to v5.0.0 to be compared with v4 as we increased the final bundle size after adding more features.

Also, I'm thinking if there's a way to decrease the final bundle without removing any feature

gabrieljablonski:
> Also, I'm thinking if there's a way to decrease the final bundle without removing any feature

I don't think there's an obvious way to reduce bundle size, since the package only has 2 actual dependencies.

PR #902, opened 2023-01-11
Improve usage guide for default css styling

gabrieljablonski:


docs/docs/examples/styling.mdx
```diff
@@ -365,7 +365,7 @@ import 'react-tooltip/dist/react-tooltip.css'
 
 #### Observation
 
-This doc is only if you are importing `import 'react-tooltip/dist/react-tooltip.css'`, if not, you can write your classes by you directly.
+This doc is only if you are importing `import 'react-tooltip/dist/react-tooltip.css'`. If not, you can write your own CSS classes directly.
```
gabrieljablonski, line 368: I believe this suggestion should be removed completely. Is it really viable for someone to write all of the CSS classes themselves? 

If so, how would they go about it? Since after bundling the class names get mangled (e.g. `.tooltip` becomes `.styles-module_tooltip__mnnfp`).

danielbarion: hmm, got your point!

They can add their own CSS by className props, the idea is to let them customize everything they want in the appearance of the tooltip.

But you're right, there's no benefit to writing the entire CSS from scratch, it's better to only override what they want to change.

I agree with you.

danielbarion (approved):

PR #907, opened 2023-01-18
fix: replace useId with v4

maosin77:
`useId` was introduced in React 18 but a lot of projects still use React 17.
to achieve backward compability we could replace useId with uuid.

This PR:
- replaces useId with uuid

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,8 @@
   },
   "dependencies": {
     "@floating-ui/dom": "^1.0.4",
-    "classnames": "^2.3.2"
+    "@types/uuid": "^9.0.0",
+    "classnames": "^2.3.2",
+    "uuid": "^9.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


gabrieljablonski:
99% of the time when using the `uuid` package, it can be safely changed to `nanoid`, which is an order of magnitude smaller when it comes to bundle size, and often achieves the same goal.

While I do agree making the package dependent on React 18 just because of one call to `useId()` may not be the best, we should find an approach that does not require adding a new dependency.

danielbarion:
Hi, I agree with @gabrieljablonski.

I don't like the idea of including `uuid` again in the project to generate the IDs, as you can see in the image, we reduced the bundle size from v4 to v5, one of the reasons is the drop of `uuid` package.

image

But I do agree that we should support React <= v18 as v4 still has a few issues and it is not maintained anymore.

gabrieljablonski:
After looking at the source code, I've realized a unique id is not required for the provider to work properly. In a previous version, a unique id was essential, so `useId()` ended up coming along it was no longer needed.

Along with some other improvements #911 drops `useId()` usage, so React 18 will no longer be required.

maosin77:
Thanks for your comments, I didn't know `nanoId`. I hope this change will bring value to others as well :)

PR #911, opened 2023-01-18
Improve internal provider api and drop usage of `useId()` (React v18 no longer required)

gabrieljablonski:
React v18 is no longer required since `useId()` was dropped.


danielbarion:
@gabrieljablonski minimum supported React version dropped to `v16.14` in package.json.

danielbarion (approved):

PR #921, opened 2023-01-27
Fix `Middleware` type self-reference on build output

### LINKED ISSUE #920: [BUG] TS compile: "TS2456: Type alias 'Middleware' circularly references itself."
**Describe the bug**
Error thrown when compiling react app with typescript v4.9.4 and using yarn pnp.
[Example project](https://github.com/bobmazy/react-typescript)

`.yarn/__virtual__/react-tooltip-virtual-7b0a038807/0/cache/react-tooltip-npm-5.7.0-8f25b21000-1d3ae9afc9.zip/node_modules/react-tooltip/dist/react-tooltip.d.ts:16:6 - error TS2456: Type alias 'Middleware' circularly references itself.`

**Version of Package**
v5.6.0

**To Reproduce**
1. add typescript `v4.9.4` to react app 
2. use yarn pnp
3. use react-tooltip `v5.6.0`
4. compile

or 
1. clone https://github.com/bobmazy/react-typescript
2. run `yarn`
3. run `yarn tsc`

**Expected behavior**
React App should compile without any error

gabrieljablonski:
Closes #920.

This seems like a bug in rollup. The output `.d.ts` file does not include the `floating-ui` import, probably because rollup gets confused by using the exact same name for both types.

```ts
// Not included in the output
// import type { Middleware } from '@floating-ui/dom'
export type Middleware = Middleware
```

### DIFF
--- a/src/components/Tooltip/TooltipTypes.d.ts
+++ b/src/components/Tooltip/TooltipTypes.d.ts
@@ -1,5 +1,6 @@
 import type { ElementType, ReactNode, CSSProperties } from 'react'
-import type { Middleware } from '@floating-ui/dom'
+
+export type { Middleware } from '@floating-ui/dom'
 
 export type PlacesType = 'top' | 'right' | 'bottom' | 'left'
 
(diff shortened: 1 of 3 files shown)


gabrieljablonski:
Apparently, this [should be failing at compilation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#local-and-imported-type-declarations-now-conflict) by TypeScript, so maybe rollup is just ignoring it.

This should be triggering error TS2440, not sure why it isn't.

danielbarion (approved):

PR #932, opened 2023-02-01
fix: remove tooltip rendered element from DOM when is not showing

### LINKED ISSUE #918: [BUG] Tooltips aren't completely hidden on page when not showing
**Describe the bug**

Tooltips are not being fully hidden in the DOM

**Version of Package**
v5.4.0

**To Reproduce**

If you have a list of items (particularly one longer than the view-height of the page), tooltips will take up space in the background even if not active, causing the actual height of the page to be greater than necessary (on our app this caused a huge block of scrollable white space underneath our app).

On inspection, I noticed that tooltips have a CSS property of `visibility: hidden` rather than `display: none`, causing it to take up space (even if invisible).

**Expected behavior**

It should be using `display: none` when the tooltip is not showing to properly remove it from page calculations.

danielbarion:
- [x] fix the tooltip `first show` when the element is not rendered (the position was miscalculated)
- [x] stay with the same behaviors in this PR as before

closes #918

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -59,19 +59,36 @@ const Tooltip = ({
 
   const handleShow = (value: boolean) => {
     setRendered(true)
-
     /**
      * wait for the component to render and calculate position
      * before actually showing
      */
     setTimeout(() => {
-      if (setIsOpen) {
-        setIsOpen(value)
-      } else if (isOpen === undefined) {
+      setIsOpen?.(value)
+      if (isOpen === undefined) {
         setShow(value)
       }
     }, 10)
   }
+
+  /**
+   * this replicates the effect from `handleShow()`
+   * when `isOpen` is changed from outside
+   */
+  useEffect(() => {
+    if (isOpen === undefined) {
+      return () => null
+    }
+    if (isOpen) {
+      setRendered(true)
+    }
+    const timeout = setTimeout(() => {
+      setShow(isOpen)
+    }, 10)
+    return () => {
+      clearTimeout(timeout)
+    }
+  }, [isOpen])
 
   useEffect(() => {
     if (show === wasShowing.current) {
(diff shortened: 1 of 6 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -386,13 +401,19 @@ const Tooltip = ({
   }, [])
 
   const hasContentOrChildren = Boolean(html || content || children)
+  const canShow = Boolean(
+    hasContentOrChildren &&
+      !calculatingPosition &&
+      (isOpen || show) &&
+      Object.keys(inlineStyles).length > 0,
```
danielbarion, line 408: `Object.keys(inlineStyles).length > 0` this code with the small delay to show the tooltip (time to get the tooltip element rendered before position calculation) fixed the problem of tooltip start showing in a wrong place for the first time

src/components/Tooltip/Tooltip.tsx
```diff
@@ -44,19 +44,34 @@ const Tooltip = ({
...
+     * this `timeout` is necessary because the component
+     * needs to be rendered to calculate the position correctly
+     */
+    setTimeout(() => {
+      if (setIsOpen) {
+        setIsOpen(value)
+      } else if (isOpen === undefined) {
+        setShow(value)
+      }
+    }, 10)
```
danielbarion, line 74: needs to double-check the delay tooltip

src/components/Tooltip/Tooltip.tsx
```diff
@@ -44,19 +44,34 @@ const Tooltip = ({
   const [inlineStyles, setInlineStyles] = useState({})
   const [inlineArrowStyles, setInlineArrowStyles] = useState({})
   const [show, setShow] = useState(false)
+  const [rendered, setRendered] = useState(false)
```
danielbarion, line 47: added a new prop to handle if the tooltip can be rendered or not, this way we can handle the animations and the element render

src/components/Tooltip/Tooltip.tsx
```diff
@@ -44,19 +44,34 @@ const Tooltip = ({
...
-  const handleShow = (value: boolean) => {
-    if (setIsOpen) {
-      setIsOpen(value)
-    } else if (isOpen === undefined) {
-      setShow(value)
+  useEffect(() => {
+    if (!show) {
+      setRendered(false)
     }
+  }, [show])
```
danielbarion, line 59: need double-check the delay tooltip

danielbarion:
Test snapshots not updated yet

src/components/Tooltip/Tooltip.tsx
gabrieljablonski: Ideally we shouldn't use a timeout, and actually wait for the styles to be updated. 

I couldn't see weird behaviors, so it should be fine for now, but it's something to think about.

gabrieljablonski:
Tests will need to be fixed, since they assumed the tooltip was always in the DOM.

After that, this should be good to merge.

gabrieljablonski (approved):
The tests are simple enough that we probably should be fine for now without them.

PR #935, opened 2023-02-03
fix: wrapper element changing after first render

### LINKED ISSUE #929: [FEAT REQ] Customizable Arrow Wrapper

When using `react-tooltip` with docusaurus, it results with unexpected behavior. This is because `react-tooltip` defaults to using `div` for the tooltip and arrow elements and it results to these `div` elements being child elements of a `p` element, which is not allowed in docusaurus.

It's possible to change the wrapper element of the tooltip using the `wrapper` prop, but it's not possible for the arrow.

Allow customizing the element used for the arrow similar to the `wrapper` prop that customizes the tooltip wrapper element.

**Describe alternatives you've considered**

When the `noArrow` prop is set to `true`, remove the arrow element instead of adding a class that hides it.

danielbarion:
close #929

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -19,7 +19,7 @@ const Tooltip = ({
   events = ['hover'],
   positionStrategy = 'absolute',
   middlewares,
-  wrapper: WrapperElement = 'div',
+  wrapper: WrapperElement,
   children = null,
   delayShow = 0,
   delayHide = 0,
@@ -38,7 +38,7 @@ const Tooltip = ({
   setIsOpen,
 }: ITooltip) => {
   const tooltipRef = useRef<HTMLElement>(null)
-  const tooltipArrowRef = useRef<HTMLDivElement>(null)
+  const tooltipArrowRef = useRef<HTMLElement>(null)
   const tooltipShowDelayTimerRef = useRef<NodeJS.Timeout | null>(null)
   const tooltipHideDelayTimerRef = useRef<NodeJS.Timeout | null>(null)
   const [inlineStyles, setInlineStyles] = useState({})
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -38,7 +38,7 @@ const Tooltip = ({
   setIsOpen,
 }: ITooltip) => {
   const tooltipRef = useRef<HTMLElement>(null)
-  const tooltipArrowRef = useRef<HTMLDivElement>(null)
```
danielbarion, line 41: updated to match other possible elements

src/components/Tooltip/Tooltip.tsx
```diff
@@ -19,7 +19,7 @@ const Tooltip = ({
   events = ['hover'],
   positionStrategy = 'absolute',
   middlewares,
-  wrapper: WrapperElement = 'div',
```
danielbarion, line 22: default value removed because there's no way to use Tooltip component directly without Controller and Controller always passes a `wrapper` to Tooltip

src/components/Tooltip/TooltipTypes.d.ts
```diff
@@ -43,7 +43,7 @@ export interface ITooltip {
   id?: string
   variant?: VariantType
   anchorId?: string
-  wrapper?: WrapperType
```
danielbarion, line 46: possible `undefined` removed because there's no way to use Tooltip component directly without Controller and Controller always passes a `wrapper` to Tooltip

src/components/TooltipController/TooltipController.tsx
```diff
@@ -85,7 +85,7 @@ const TooltipController = ({
         setTooltipOffset(value === null ? offset : Number(value))
       },
       wrapper: (value) => {
-        setTooltipWrapper((value as WrapperType) ?? 'div')
```
danielbarion, line 88: the fix.

when the component start to check anchor attributes and there's no `wrapper`, the `wrapper` state gets updated to be a `div`

gabrieljablonski (approved):

PR #940, opened 2023-02-10
fix(TooltipProvider): add TooltipProviderProps

joshuayoes:
react 17+ types require a generic argument to be passed to PropsWithChildren, even if it has no other props


src/components/TooltipProvider/TooltipProvider.tsx
```diff
@@ -34,7 +34,9 @@ const DEFAULT_CONTEXT_DATA_WRAPPER: TooltipContextDataWrapper = {
 
 const TooltipContext = createContext<TooltipContextDataWrapper>(DEFAULT_CONTEXT_DATA_WRAPPER)
 
-const TooltipProvider: React.FC<PropsWithChildren> = ({ children }) => {
+interface TooltipProviderProps {}
+
+const TooltipProvider: React.FC<PropsWithChildren<TooltipProviderProps>> = ({ children }) => {
```
gabrieljablonski, line 39:
```suggestion
const TooltipProvider: React.FC<PropsWithChildren<void>> = ({ children }) => {
```

Try doing `void` or `unknown` (preferably `void`) instead and let me know if the compiler or something else complains about it.
The linter doesn't like empty interfaces, so let's try to keep it more concise.

gabrieljablonski (changes requested):

gabrieljablonski (approved):

PR #948, opened 2023-02-17
Fix some bugs and general tooltip behavior

gabrieljablonski:
- [X] `setRendered(false)` when anchor element gets removed from the DOM
- [X] `setShow(true)` only after finishing calculating position
- [X] Wait to close the tooltip so it doesn't fade back in when quickly switch between anchors
- [ ] Fix behavior when using `isOpen`/`setIsOpen`

When using timeout, the tooltip might still get shown before calculating the position.
Instead of using a timeout, render the tooltip and wait for the position calculation to finish.


gabrieljablonski:
Couldn't get some stuff to work exactly as it should, and this would probably break other stuff as well.

Might try something like this again in the future.

PR #952, opened 2023-02-19
Watch for added nodes and include in anchors list

### LINKED ISSUE #951: [BUG] v5.8.1 Not properly linking dynamically loaded triggers when one does not exist on mount
**Describe the bug**
@gabrieljablonski  Just a follow-up issue on v5.8:

There is an issue when using `data-tooltip-id`, if a `<Tooltip/>` exists on the initial load, but a triggering element _does not_ exist on the initial load; When trigger elements become available in a subsequent view, they will not trigger the connection with the initial loaded `<Tooltip/>`. 

**Version of Package**
v5.8.1

**To Reproduce**
Live example can be seen here: https://stackblitz.com/edit/react-dhtxpv?file=src%2FApp.js

**Expected behavior**
Regardless of whether a triggering element exists on the initial load of the page, the connection should be maintained.

**Desktop**
 - OS: iOS
 - Browser: Chrome
 - Version: 110
 - Frameworks: React 18
 
**Additional context**
**Note:** If even just a single, dummy, triggering element (eg. `<span data-tooltip-id="tooltip" data-tooltip-content="." />`) is added to the DOM on initial load (but only inside a conditional statement, and not the immediate DOM? See example comments...), it actually resolves this issue with subsequent views.

gabrieljablonski:
Closes #951

### DIFF
--- a/src/test/tooltip-props.spec.js
+++ b/src/test/tooltip-props.spec.js
@@ -130,8 +130,6 @@ describe('tooltip props', () => {
       },
     )
 
-    await userEvent.unhover(anchorElement)
-
     await waitFor(() => {
       tooltip = screen.getByRole('tooltip')
     })
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -359,7 +359,14 @@ const Tooltip = ({
       selector = `[data-tooltip-id='${id}']`
     }
     const documentObserverCallback: MutationCallback = (mutationList) => {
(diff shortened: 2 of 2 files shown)


src/test/tooltip-props.spec.js
```diff
@@ -130,8 +130,6 @@ describe('tooltip props', () => {
       },
     )
 
-    await userEvent.unhover(anchorElement)
```
gabrieljablonski, line 133: How was this test passing before? 🤔

danielbarion (approved):

PR #957, opened 2023-02-22
Remove floating-ui exports to improve bundlesize

gabrieljablonski:
We don't even have it on the documentation that you can import the middlewares directly from the package.
If needed, the middlewares can be imported directly from `@floating-ui/dom`.


danielbarion (approved):
It makes sense to me, thanks!

PR #958, opened 2023-02-22
Add external deps rollup plugin

gabrieljablonski:
Adding the plugin without with default configurations seems to work fine.

Is there a reason we shouldn't do this?

Sample project: https://stackblitz.com/edit/react-ts-22ya1r?file=App.tsx,package.json
Bundlephobia: https://bundlephobia.com/package/react-tooltip@5.8.2-beta.4

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "rollup-plugin-dts": "^5.0.0",
     "rollup-plugin-filesize": "^9.1.1",
     "rollup-plugin-html-scaffold": "^0.2.0",
+    "rollup-plugin-peer-deps-external": "^2.2.4",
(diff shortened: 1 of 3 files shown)


danielbarion:
When I did a quick test for this change, the package got broken, unsure if the reason was a bug or package misconfiguration.

Was this tested? I propose to we release a beta version to be sure this will work as expected.

danielbarion:
The Esbuild PR: https://github.com/ReactTooltip/react-tooltip/pull/959
image

https://bundlephobia.com/package/react-tooltip@5.8.4-beta.2

gabrieljablonski:
Using esbuild instead #959

PR #959, opened 2023-02-22
feat: add esbuild as bundler

danielbarion:
- [x] Add Esbuild to the project
- [x] Make the development build with esbuild work like the development build with rollup
- [x] Make the production build with esbuild work like the production build with rollup
- [ ] ~Try to generate the types with Esbuild~ - directly is not possible
- [x] Check the filesize of the files to see if the esbuild is lighter than rollup generated files
- [x] Check if all examples in docs are working as expected
- [x] Check if tests are working as expected
- [x] Add live reload to dev environment
----
- [x] Revert React version to use `16` instead of `18` 
- [x] Remove **create-react-app** (`cra-app`) folder when it's not used anymore
----

https://npmtrends.com/esbuild-vs-rollup-vs-snowpack-vs-swc-vs-vite


.gitignore
```diff
@@ -35,3 +35,7 @@ yarn-error.log
 !.babelrc
 !.gitignore
 !.npmignore
+
+# yalc
+/.yalc
+
```
danielbarion, line 41: I was testing the dependencies using `yalc` (https://github.com/wclr/yalc)

cra-app/src/App.js
```diff
@@ -0,0 +1,30 @@
...
+          className="App-link"
+          href="https://reactjs.org"
+          target="_blank"
+          rel="noopener noreferrer"
+          data-tooltip-id="my-tooltip"
+          data-tooltip-content="Hello world!"
+        >
+          Learn React
+        </a>
+        <Tooltip id="my-tooltip" />
```
danielbarion, line 24: Rendering the tooltip on CRA

docs/.gitignore
```diff
@@ -18,3 +18,8 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+
+# yalc
+/.yalc
+yalc.lock
```
danielbarion, line 25: I was testing the dependencies using `yalc` (https://github.com/wclr/yalc)

docs/package.json
```diff
@@ -21,9 +21,9 @@
     "clsx": "^1.2.1",
     "prism-react-renderer": "^1.3.5",
     "raw-loader": "^4.0.2",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
-    "react-tooltip": "5.8.1"
+    "react": "link:../node_modules/react",
+    "react-dom": "link:../node_modules/react-dom",
+    "react-tooltip": "link:.."
```
danielbarion, line 26: Those changes need to be reverted before the PR is merged

esbuild.config.dev.mjs
```diff
@@ -0,0 +1,32 @@
...
+
+await ctx.watch()
+
+const { port } = await ctx.serve({
+  servedir: 'build',
+  port: 3000,
+  host: 'localhost',
+})
+
+console.log(`\n\n# Live Server: localhost:${port}`)
```
danielbarion, line 32: this config file probably can be better organized, I just tried to make the development build work as rollup one

esbuild.config.prod.mjs
```diff
@@ -0,0 +1,83 @@
...
+
+// Remove all unecessary or duplicated files from `dist` folder after build
+fs.unlink('./dist/react-tooltip.cjs.css', () => null) // generated by cjs build
+fs.unlink('./dist/react-tooltip.esm.css', () => null) // generated by esm build
+fs.unlink('./dist/react-tooltip.iife.css', () => null) // generated by iife build
+fs.unlink('./dist/react-tooltip.cjs.min.css', () => null) // generated by minified cjs build
+fs.unlink('./dist/react-tooltip.esm.min.css', () => null) // generated by minified esm build
+fs.unlink('./dist/react-tooltip.iife.min.css', () => null) // generated by minified iife build
+fs.unlink('./dist/react-tooltip.js', () => null) // generated by css build
+fs.unlink('./dist/react-tooltip.min.js', () => null) // generated by minified css build
```
danielbarion, line 83: this config file probably can be better, I just tried to make the production build work as rollup one

PS: with rollup, we export `umd`, but with esbuild we are exporting `iife`.

package.json
```diff
@@ -79,8 +84,8 @@
     "prettier": "^2.4.1",
     "process": "^0.11.10",
     "prop-types": "^15.7.2",
-    "react": "16.14.0",
-    "react-dom": "16.14.0",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
```
danielbarion, line 88: this change should be reverted before the PR is merged

package.json
```diff
@@ -3,14 +3,17 @@
...
+    "build-rollup": "node ./cli.js --env=production && npm run types && node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.prod.js && npm run bundlesize",
+    "dev": "node ./cli.js --env=development && node ./esbuild.config.dev.mjs",
+    "build": "node ./cli.js --env=production && node ./esbuild.config.prod.mjs",
     "types": "node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.types.js",
     "eslint": "eslint --ext=js --ext=jsx --ext=ts --ext=tsx --fix ./src",
     "stylelint": "stylelint \"src/**/*.css\"",
     "prettier": "prettier --config ./.prettierrc.json --write \"src/**/*{.js,.jsx,.ts,.tsx,.css}\"",
     "prepare": "husky install",
     "bundlesize": "bundlesize",
+    "esbuild": "esbuild",
```
danielbarion, line 16: not sure if this one still necessary, I added while doing the tests (before create the config file)

package.json
```diff
@@ -3,14 +3,17 @@
   "version": "5.8.1",
   "description": "react tooltip component",
   "scripts": {
-    "dev": "node ./cli.js --env=development && node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.dev.js --watch",
-    "build": "node ./cli.js --env=production && npm run types && node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.prod.js && npm run bundlesize",
+    "dev-rollup": "node ./cli.js --env=development && node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.dev.js --watch",
+    "build-rollup": "node ./cli.js --env=production && npm run types && node --max_old_space_size=2048 ./node_modules/rollup/dist/bin/rollup -c rollup.config.prod.js && npm run bundlesize",
+    "dev": "node ./cli.js --env=development && node ./esbuild.config.dev.mjs",
+    "build": "node ./cli.js --env=production && node ./esbuild.config.prod.mjs",
```
danielbarion, line 9: Just keeping two bundlers available because of comparison tests

danielbarion: also, using the CLI because esbuild doesn't clear the build/dist folder, only updates the existent files, so, the manual file to keep the folder always being cleared.

PS: I think we should rename this file, a better name than `cli.js`

danielbarion:
@gabrieljablonski 

We can test this guy which uses Esbuild to build the package (to build the types and the distribution files in one build)
https://www.npmjs.com/package/estrella

r3m0t:
Can you add source maps at the same time? I'm trying to debug a tooltip that isn't showing up.

danielbarion:
@r3m0t source map added to the build (only for this PR at the moment)

danielbarion:
image

esbuild.config.prod.mjs
```diff
@@ -0,0 +1,92 @@
...
+  builds.push(
+    esbuild.build({
+      entryPoints: ['./src/index.tsx'],
+      bundle: true,
+      outfile,
+      format,
+      treeShaking: true,
+      minify,
+      sourcemap: true,
+      external: ['react', 'react-dom', 'prop-types'],
```
gabrieljablonski, line 60: Isn't `prop-types` just a dev dependency? Does something break if we remove it as an axternal dependency?

esbuild.config.prod.mjs
```diff
@@ -0,0 +1,92 @@
...
+          v2CssModulesOption: {
+            pattern: `react-tooltip__[local]_[hash]`,
+          },
+        }),
+      ],
+    }),
+  )
+})
+
+await Promise.all(builds)
```
gabrieljablonski, line 74: `Promise.allSettled()` would fail silently if any errors ocurred.

esbuild.config.prod.mjs
gabrieljablonski: There's probably a way to disable generating the CSS for some builds, and have just one. Couldn't find it though, so I'm fine with leaving this as is.

gabrieljablonski (approved):

esbuild.config.prod.mjs
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * delete the extra build files from the CSS build
+ */
+toDelete.add('dist/react-tooltip.js')
+toDelete.add('dist/react-tooltip.js.map')
+
+toDelete.forEach((file) => {
+  fs.unlink(file, () => null)
+})
```
gabrieljablonski, line 97: This might be a better way to do a postbuild cleanup, since we won't have to add/remove unlinks when adding/removing build formats.

esbuild.config.prod.mjs
danielbarion: That's true!

Thanks for this fix!

gabrieljablonski (approved):

PR #970, opened 2023-03-06
Expose tooltip placement using classes

### LINKED ISSUE #969: [FEAT REQ]: Indicate the actual place where tooltip is rendered in some way
I have a similar need to [this feature request](https://github.com/ReactTooltip/react-tooltip/issues/868), but I don't think there have been any changes made related to it before it was closed.

I can use the `place` prop that is being set on the tooltip in my styling to "track" the place that is being set, but the `place` I am setting is not always respected. For example, if I have set `place="top"`, but there is no room at the top of the anchor element, the tooltip is positioned at the bottom of the anchor element. I can see why this is done, but the styling I have depending on `place` has no knowledge of the placement being different than the prop value, so the styling used is different than what it should be for the actual placement when it differs from the set value.

From the place prop description: `Position relative to the anchor element where the tooltip will be rendered (if possible).` The "if possible" part is what makes it impossible to use the same value for `place` in my implementation of this tooltip.

Can we expose the `place` context or apply classNames to the tooltip to reflect the actual placement used?

**Additional context**
I am adding a border to the arrow shape and need to use the placement to determine what sides of the border should be set to `none` in order to only show a border around the outer edge.

gabrieljablonski:
Closes #969 

Beta release for testing: `yarn add react-tooltip@5.9.1-beta.0`

### DIFF
--- a/src/test/__snapshots__/tooltip-props.spec.js.snap
+++ b/src/test/__snapshots__/tooltip-props.spec.js.snap
@@ -27,7 +27,7 @@ exports[`tooltip props clickable tooltip 1`] = `
     Lorem Ipsum
   </span>
   <div
-    class="react-tooltip undefined react-tooltip__place-top"
+    class="react-tooltip react-tooltip__place-top undefined"
     role="tooltip"
     style="left: 5px; top: -10px;"
   >
--- a/src/test/__snapshots__/tooltip-props.spec.js.snap
+++ b/src/test/__snapshots__/tooltip-props.spec.js.snap
@@ -8,7 +8,7 @@ exports[`tooltip props basic tooltip 1`] = `
     Lorem Ipsum
   </span>
   <div
-    class="react-tooltip"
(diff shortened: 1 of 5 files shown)


src/test/__snapshots__/tooltip-props.spec.js.snap
```diff
@@ -27,7 +27,7 @@ exports[`tooltip props clickable tooltip 1`] = `
     Lorem Ipsum
   </span>
   <div
-    class="react-tooltip undefined"
+    class="react-tooltip react-tooltip__place-top undefined"
```
gabrieljablonski, line 30: when using the `clickable` prop, we get an `undefined` classname when testing. this doesn't happen outside of testing.

any ideas @danielbarion?

danielbarion: that's weird, I'll take a look when I have some available time, but this is not a blocker for this PR as this is an `extra-class`

PR #988, opened 2023-03-22
fix: update node contains conditional check on Tooltip component

### LINKED ISSUE #985: Error when calling node.contains because it is not a method
https://github.com/ReactTooltip/react-tooltip/blob/master/src/components/Tooltip/Tooltip.tsx#L377

When `node` is undefined or when `contains` is not a method this causes an uncaught exception.

Suggest changing the expression to:

```
node?.contains && node.contains(activeAnchor)
```

This will prevent errors from happening

danielbarion:
closes #985

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -374,7 +374,7 @@ const Tooltip = ({
         }
         if (activeAnchor) {
           ;[...mutation.removedNodes].some((node) => {
-            if (node.contains(activeAnchor)) {
+            if (node?.contains && node.contains(activeAnchor)) {
               setRendered(false)
               handleShow(false)
               setActiveAnchor(null)
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -374,7 +374,7 @@ const Tooltip = ({
         }
         if (activeAnchor) {
           ;[...mutation.removedNodes].some((node) => {
-            if (node?.contains && node.contains(activeAnchor)) {
+            if (node?.contains?.(activeAnchor)) {
               setRendered(false)
               handleShow(false)
               setActiveAnchor(null)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -374,7 +374,7 @@ const Tooltip = ({
         }
         if (activeAnchor) {
           ;[...mutation.removedNodes].some((node) => {
-            if (node.contains(activeAnchor)) {
+            if (node?.contains && node.contains(activeAnchor)) {
```
gabrieljablonski, line 377:
```suggestion
            if (node?.contains?.(activeAnchor)) {
```

gabrieljablonski (changes requested):

gabrieljablonski:
I have no problem with doing this change but it feels unnecessary. `mutation.removedNodes` should never have a `null` node inside it, and if a node is not `null`, it will always have the `contains()` method defined.

Maybe in older browsers something weird might happen? Or maybe if the reference to the node becomes stale `contains` becomes `null`?

Again, we can do the change just in case, but if #985 wasn't just a suggestion, and @rolandpoulter actually run into an error because of this, I'd love to see a reproducible example purely out of curiosity.

danielbarion:
@gabrieljablonski I do agree with you, can you handle this PR and this issue, please?

rolandpoulter:
@gabrieljablonski I can look into the reason why, but I did get an error after upgrading to the latest version. I patched it in my node_modules with the above change because I noticed in the code you are checking for node to be truthy before calling contains elsewhere. There might be something on my end with how I wrapped the tooltip that causes this issue but it will take some time to investigate and provide the exact reproduction.

gabrieljablonski:
> @gabrieljablonski I can look into the reason why, but I did get an error after upgrading to the latest version. I patched it in my node_modules with the above change because I noticed in the code you are checking for node to be truthy before calling contains elsewhere. There might be something on my end with how I wrapped the tooltip that causes this issue but it will take some time to investigate and provide the exact reproduction.

@rolandpoulter No worries, just wanted to make sure this was what solved your problem. We'll merge this change and release a new version soon.

Thanks for reporting it!

gabrieljablonski (approved):

PR #994, opened 2023-03-28
Add `styles` and `render` as dependencies for the position calculation effect

### LINKED ISSUE #993: Position is off if the content of the tootip's height changes
### Discussed in https://github.com/ReactTooltip/react-tooltip/discussions/991

<div type='discussions-op-text'>

<sup>Originally posted by **akraines** March 28, 2023</sup>
It seems that position is off if the content of the tooltip's height changes. Is there a way to indicate to the tooltip to recalculate its position based on the height of the dynamic content of the tooltip?
I'm trying to work out why tooltips are being positioned way out of place on first render of the cell in my app. On the second render it positions itself correctly. I have considered 2 possibilities:
1. The tooltip is associated with a row that is being rendered as part of a virtual list.
2. The height of the tooltip changes as part of its first render [it first says loading and then displays the content.] As a test I gave the tooltip a fixed height and it seemed to fix the issue.</div>

I looked into it further and made a repro: https://stackblitz.com/edit/react-ts-ay7kkp?file=App.tsx,index.tsx

gabrieljablonski:
Closes #993.

A down side might be that if styles is passed inline like this:

```tsx
<Tooltip
  styles={{
    attr: value
  }}
/>
```

since the `styles` object will change on every render, the tooltip will recalculate its position on every render. 

That doesn't seem like a big deal, but is there a smart way to avoid this, besides forcing the user to wrap the object with `useMemo()` or something similar?

----

https://stackblitz.com/edit/react-ts-rtgmca

- [X] add `styles` as a dependency for the calculation
- [X] watch for attribute changes on the component passed through the `render` prop

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -34,7 +34,7 @@ const Tooltip = ({
   afterHide,
   // props handled by controller
   content,
-  contentRef,
+  contentWrapperRef,
   isOpen,
   setIsOpen,
   activeAnchor,
@@ -476,21 +476,17 @@ const Tooltip = ({
   }, [show, activeAnchor, content, externalStyles, place, offset, positionStrategy, position])
 
   useEffect(() => {
-    if (!contentRef?.current) {
+    if (!contentWrapperRef?.current) {
       return () => null
     }
-    const contentObserver = new MutationObserver(() => {
+    const contentObserver = new ResizeObserver(() => {
       updateTooltipPosition()
     })
-    contentObserver.observe(contentRef.current, {
-      attributes: true,
-      childList: true,
-      subtree: true,
-    })
+    contentObserver.observe(contentWrapperRef.current)
     return () => {
       contentObserver.disconnect()
     }
-  }, [content, contentRef?.current])
+  }, [content, contentWrapperRef?.current])
 
   useEffect(() => {
     const anchorById = document.querySelector<HTMLElement>(`[id='${anchorId}']`)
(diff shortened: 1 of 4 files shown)


akraines:
use a structural compare?

gabrieljablonski:
Test with `react-tooltip@5.10.2-beta.3`: https://stackblitz.com/edit/react-ts-rtgmca

This solution feels a bit cumbersome with having to forward the ref to the custom component, but not sure if there are any other viable ways of achieving this.

akraines:
Worked like a charm - thatnks for fixing this so quickly!
I think that the ref approach makes alot of sense - I was actually thinking of recommending using react-measure.
 In fact, that might still be a good idea - this way you could just wrap the result of the render with a measure [or your own resize observer like you did here] and then the user wont need to pass in the ref.

gabrieljablonski:
> you could just wrap the result of the render with a measure [or your own resize observer like you did here] and then the user wont need to pass in the ref.

that's actually a great idea. I'll try it out whenever i get the time

gabrieljablonski:
With `react-tooltip@5.10.4-beta.994.1` no `ref` is required: https://stackblitz.com/edit/react-ts-rtgmca

@akraines thanks for the great tips. I believe using the wrapper shouldn't be a problem in 99% of cases, we'll see what we can do if it ever causes any issues.

Also you mentioning `react-measure` led me to the `ResizeObserver` API, which I forgot existed, and is a lot more elegant to use in this case over the `MutationObserver` API. So thanks again!

As soon as @danielbarion reviews it, we'll merge and do an official release.

danielbarion (approved):
Code looks good to me :)

akraines:
Nice and clean. Thank you very much for taking the time to fix this

PR #1017, opened 2023-05-01
Fixed the `id` in the clickable tooltip example

MosheEichler:
The id of the clickable example should be `my-tooltip-click` not `my-tooltip`


danielbarion:
Hi @MosheEichler, thanks for the PR!

danielbarion (approved):

PR #1026, opened 2023-05-19
feat: log invalid CSS selector with tooltip-id

dtamai:
## Context

There is this warning for when the CSS selector is invalid in TooltipController.

It was useful for me when I troubleshooting an issue with a tooltip but it would print only `[react-tooltip] "undefined" is not a valid CSS selector`, but the issue was that the CSS selector contained a `\r\n` that made it invalid, not that it was undefined.

The warning is not printing the invalid selector when it is using the `data-tooltip-id` attribute. This PR proposes to print the `selector` variable that covers both selectors - anchor and id.

This produces the warning below (there is a line break in the selector):

image

---

:tada: And thanks for maintaining this package! The recent addition of the try/catch block that prints the warning means that only the tooltip breaks, not the whole page.

### DIFF
--- a/src/components/TooltipController/TooltipController.tsx
+++ b/src/components/TooltipController/TooltipController.tsx
@@ -182,7 +182,7 @@ const TooltipController = ({
       } catch {
         if (!process.env.NODE_ENV || process.env.NODE_ENV !== 'production') {
           // eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


danielbarion (approved):

danielbarion:
@dtamai thanks for the fix!

Issue #1025, opened 2023-05-18
[BUG] classNames are ignored in v5.13.0


ansgarsteinkamp:
Bug in v5.13.0: classNames like the Tailwind-Class `bg-blue-900` are ignored somehow:

```
<Tooltip className="z-50 bg-blue-900" id="Home" delayShow={200}>
   Home
</Tooltip>
```

With v5.12.0 everything works just fine.

danielbarion:
Hi, I tested here and the issue is:

* CSS is being injected into the page

CSS works like objects, when we pass a custom class, this is the behavior:

Before `v5.13.0`

1. import CSS styles:
2. component receive a new class
3. browser process and render it

```
.tooltip-default-css {
  background: red;
}

.bg-green-500 {
  background: green;
}

--------- same as:

Object.assign({ background: 'red' }, { background: 'green' })
```

On version `v5.13.0`:

1. component receives a new class
2. CSS is injected into the page
4. browser process and render it

```
.bg-green-500 {
  background: green;
}

.tooltip-default-css {
  background: red;
}
--------- same as:

Object.assign({ background: 'green' }, { background: 'red' })
```


this is a `bug`, we will take a look as soon as possible. We are open to suggestions on fixing this because the CSS class from props and injected CSS have the same CSS specificity.

Solution 1 -> you need to increase the CSS specificity from your side - I don't like to force the community using our library to write extra code like this
Solution 2 -> we need to increase the specificity of the CSS received by props (how?) - I prefer this approach, but I don't have any ideas on how to do that at the momentn


reference: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

gabrieljablonski:
#1027 suggests a workaround

danielbarion:
yeah, that workaround is just the dev increasing the specificity of his CSS query, it's the solution 1 that I mentioned.

another way to do that together with taiwlind is:

```
HTML - JSX
<div className="container">
  <Tooltip className="tooltip" />
</div>

CSS
.container .tooltip {
  @apply bg-green-500;
}
```

Tomekmularczyk:
Hey guys, I don't think that's problem with the current version. Tailwind classes were ignored with react-tooltip@5.8.3. This is why I had to create separate CSS class like:

```css
.tooltipWrapper { 
  @apply rounded-lgplus bg-tixy-1000 p-2 text-sm font-normal leading-4 text-tixy-10 shadow-sm;
}
```

[...]

PR #1041, opened 2023-06-14
feat: add manual inject styles function into the project

danielbarion:
idea from #1025


danielbarion:
I created the flow to inject the styles manually from the package instead of letting Rollup handle it by itself.

what is pending?
* add a switch to enable and disable styles injection.

Suggestion: 
In the current way that this was implemented, we can add the switch by environment variables, ex:
`REACT_TOOLTIP_DISABLE_STYLES=true`

But this doesn't sound like the best way to handle this switch.

So, I need help here :)

src/utils/handle-style.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+function removeStyle() {
+  const style = document.getElementById(REACT_TOOLTIP_STYLES_ID)
+  if (!style) {
+    return
+  }
+  style.remove()
+}
```
gabrieljablonski, line 44: Please test if something like this works (too lazy to setup and test it myself 😑)

The idea is to use `import { removeStyle } from 'react-tooltip'`

src/index.tsx
```diff
@@ -13,6 +16,11 @@ import type {
 import type { ITooltipController } from './components/TooltipController/TooltipControllerTypes'
 import type { ITooltipWrapper } from './components/TooltipProvider/TooltipProviderTypes'
 
+// this content will be replaced in build time with the `react-tooltip.css` builded content
+const TooltipStyles = 'react-tooltip-css-placeholder'
+
+injectStyle(TooltipStyles)
```
gabrieljablonski, line 22: I didn't mean to commit it, but why was `@ts-ignore` being used here? No warnings or errors on my environment.

src/utils/handle-style.ts
```diff
@@ -0,0 +1,46 @@
...
+    style.appendChild(document.createTextNode(css))
+  }
+}
+
+function removeStyle() {
+  const style = document.getElementById(REACT_TOOLTIP_STYLES_ID)
+  if (!style) {
+    return
+  }
+  style.remove()
```
gabrieljablonski, line 43: Also if you make any changes, do this one as well, forgot to clean it up.

```suggestion
  const style = document.getElementById(REACT_TOOLTIP_STYLES_ID)
  style?.remove()
```

gabrieljablonski:
How does removing the style injection affect the tooltip functionally? Since it depends on the `.tooltip` and `.show` classes to work properly, will the tooltip still show and hide without any issues?

danielbarion:
> How does removing the style injection affect the tooltip functionally? Since it depends on the `.tooltip` and `.show` classes to work properly, will the tooltip still show and hide without any issues?

This will probably break the tooltip feature, one thing that we can do is separate it into 2 CSS files:
1. for basic functionality - keep the tooltip working as expected
2. for styling - without this one, we will only show and hide the tooltip content without any styling

please let me know your thoughts about it

src/index.tsx
danielbarion: yeap, all good! thanks!

gabrieljablonski:
@danielbarion Having a separate injection for functional styling seems like an adequate solution.

My suggestion (make sure to check if I missed something):

```css
/* react-tooltip-core.css (feel free to choose a better name for the file) */
.tooltip {
  position: absolute;
  visibility: hidden;
  opacity: 0;
}

.fixed {
  position: fixed;
}

.arrow {
  position: absolute;
}

.noArrow {
  display: none;
}

.show {
  visibility: visible;
  opacity: var(--rt-opacity);
}
```

`.fixed` and `.noArrow` could be left out. I included them just so the `positionStrategy` and `noArrow` props still work. Your call if we should consider them functional styling or not.

danielbarion:
Beta version released with the last commit:

```
yarn add react-tooltip@5.15.0-beta.1041.0
```
or
```
npm install react-tooltip@5.15.0-beta.1041.0
```

I only tested without the styles, I'll test everything later, but feel free to test it or update anything.

Thanks!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -561,28 +562,34 @@ const Tooltip = ({
...
-          /**
-           * changed from dash `no-arrow` to camelcase because of:
-           * https://github.com/indooorsman/esbuild-css-modules-plugin/issues/42
-           */
-          [styles['noArrow']]: noArrow,
-        })}
+        className={classNames(
+          'react-tooltip-arrow',
+          coreStyles['arrow'],
+          styles['arrow'],
```
gabrieljablonski, line 583: Is keeping both styles intentional?

danielbarion:
Yes, if you check, they has different attributes on their files (if I'm not
wrong)

I will run more tests, I had to leave to computer, so, this is just to
validate the idea

Em sex., 16 de jun. de 2023 15:53, Gabriel Jablonski <
***@***.***> escreveu:

src/components/Tooltip/Tooltip.tsx
danielbarion: yes, but please feel free to reorganize the styles, I tried to keep the necessary CSS for position and basic behavior into the Core file, I'll check again this week when I have more time, but please, go ahead if you want to change anything

danielbarion:
https://github.com/ReactTooltip/react-tooltip/assets/9615850/22de8b3d-59f3-4b9b-9342-3ac9e4ab4051

@gabrieljablonski 🔝

gabrieljablonski:
> The code can be used in a button or inside of a `useEffect`

Does calling it globally (outside any components) work? That would be the most convenient IMO

danielbarion:
@gabrieljablonski I'm not totally sure that I got how to test it on Next 13, but it breaks because I'm importing from a custom export:

```
"use client"

export * from 'react-tooltip'
```

and importing this function from the package also breaks it too.

Can you test it or guide me on how to test it, please?

My test environment is just a fresh Next 13 project and I'm using Yalc to test the implementation, but you can try the beta release from NPM instead of Yalc

danielbarion:
@gabrieljablonski another option is:
* let the guys import the `styles` (basically the colors - content from styles.module.css) and we only inject the core styles by default

danielbarion:
> If we do this we're back to how it was before, which would not be ideal IMO.

I do agree, but now we have the basic feature of the Tooltip with the Core styles before nothing was working because of 0 styles of tooltip on the page. I'm just thinking of projects that don't want to inject the colors from react-tooltip and this can cause re-paint of the elements in the page (probably this impact will be minimal, but I'm thinking if it's possible to reduce it to 0).

Does it make sense?

gabrieljablonski:
Since most users will probably just use the default styling and partially override the styling, I really like the idea of not having to import something to apply the styles.

I'm fine with the env variable approach, but then we should consider how useful still having `removeStyle()` would be. Would anyone really want to remove the default styling on the fly? So having just the one way to do it feels a lot cleaner.

danielbarion:
Keeping the `removeStyles` is great for simple projects using pure HTML (importing React from CDN), or projects using other languages on the server side (like Laravel projects or something like that).

I don't know, I'm probably overthinking here, if you believe it's not useful to keep the remove styles function, I'll agree with you and all good :)

danielbarion:
@gabrieljablonski the `force push` was because I did `git pull -r origin master`, after the rebase, I had to push -f

gabrieljablonski (approved):

danielbarion:
```
yarn add react-tooltip@5.16.0-beta.1041.0
```
or
```
npm install react-tooltip@5.16.0-beta.1041.0
```

Check the PR changes for more information.

dsternlicht:
@danielbarion any idea when this is going to be merged officially? :)

gabrieljablonski:
@dsternlicht just working out the final kinks. An official version should be released in a couple days.

Until then, don't worry about using the beta version, it should work just fine.

dsternlicht:
> @dsternlicht just working out the final kinks. An official version should be released in a couple days.
> 
> Until then, don't worry about using the beta version, it should work just fine.

Sounds good. Thanks @gabrieljablonski!

danielbarion:
@gabrieljablonski I tested again and I did two new small commits, please check them before we merge this PR

thanks!

gabrieljablonski (approved):

dsternlicht:
🎉

danielbarion:
available at [v5.16.0](https://github.com/ReactTooltip/react-tooltip/releases/tag/v5.16.0)

PR #1043, opened 2023-06-20
Clear show/hide timeouts when the anchor is unmounted

### LINKED ISSUE #1042: [BUG] Tooltip displayed for unmounted component.
**Describe the bug**
When an element is set to display a tooltip (e.g. using `data-tooltip-id`), and that tooltip has a `delayShow` (e.g. 300ms), when the user triggers the tooltip (by e.g. hovering over the element), but the element is unmounted before the `delayShow` interval (e.g. after ~200ms), the tooltip for another element (always seems to be the element which is declared first) will be displayed.

**Version of Package**
v5.14.0

**To Reproduce**
See repro repo: https://github.com/knubie/react-tooltip-demo
It seems that the mouse should not move before the "phantom" tooltip is displayed.

**Expected behavior**
When the element is unmounted, no tooltip should be displayed.

**Screenshots**
https://github.com/ReactTooltip/react-tooltip/assets/431251/8b46a8c6-8f19-4d69-9451-8cc07022efed

 - OS: macOS
 - Browser firefox
 - Version 114
 - Frameworks React 18.2

gabrieljablonski:
Closes #1042.

Beta version `react-tooltip@5.14.1-beta.1043.0`

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -276,6 +276,9 @@ const Tooltip = ({
       return
     }
     handleShow(false)
+    if (tooltipShowDelayTimerRef.current) {
+      clearTimeout(tooltipShowDelayTimerRef.current)
+    }
   }
 
   const handleEsc = (event: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


danielbarion (approved):
nice find!

danielbarion (approved):

PR #1045, opened 2023-06-21
`closeOnScroll` improvements

### LINKED ISSUE #1044: [BUG] Tooltip position not updated on scroll
**Describe the bug**
Using a minimal setup like:

      <div className={"h-full overflow-scroll flex"}>
             <a className={"relative top-1/2"} data-tooltip-id="my-tooltip" data-tooltip-content="Hello world!">
                    ◕‿‿◕
               </a>
            </div>
      <Tooltip id="my-tooltip" closeOnScroll={true}/>

The position of the tooltip is not updated during scrolling. The problem can be seen here:
tooltip

The solution is to remove either "overflow:scroll" or "display:flex" from the parent div of the anchor element.
If both are present, the bug will occur. It also does not make a difference where the Tooltip is placed (in which div/DOM element), as long as the anchor parent has both CSS properties, the bug will occur.

**Version of Package**
v5.14.0

**To Reproduce**
Use the code above.

**Expected behavior**
´The tooltip disappears (closeOnScroll==true) or at least it updates the position to follow the anchor element.

 - OS: Win
 - Browser: Chrome
 - Frameworks: Next.js latest stable

gabrieljablonski:
Closes #1044.

- Ignore `delayHide` and close immediately
- Add listeners to tooltip and anchor parents

Beta version `react-tooltip@5.15.0-beta.1045.0`


gabrieljablonski:
@danielbarion if you can, please take like 30s to think about the topic order on the troubleshooting section. The current order is what I guess would be most likely of happening, to least likely.

docs/docs/troubleshooting.mdx
```diff
@@ -169,15 +169,15 @@ To use `react-tooltip` on Next.js 13 without having to tag your component or pag
...
+Avoid naming the file `react-tooltip.tsx` (or with whichever extension your project uses), since it may interfere with your editor's autocomplete functionality.
 
 :::
 
 ```jsx
 // src/components/ReactTooltip.tsx
 'use client'
 
-export * from 'react-tooltip'
+export { Tooltip } from 'react-tooltip'
```
gabrieljablonski: https://nextjs.org/docs/messages/export-all-in-page

danielbarion (approved):

danielbarion (approved):

PR #1048, opened 2023-06-25
Add "border" option including arrow

MoveRoad:
Before I write about PR, thanks for providing useful open source.

Reason for request :
I thought it would be nice to have a border option that applies to both tooltip and arrow.

As follows :
image


gabrieljablonski:
Thanks for the contribution. At first glance, everything looks ok, just gotta update the docs with the new prop. Feel free to do it yourself, or leave it to us.

We'll be reviewing this soon to see if anything should be changed, and if everything is working correctly.
Thanks again.

MoveRoad:
@gabrieljablonski Thanks for the reply. I added the new prop to docs. But I'm not good at expressing English, so you can write it yourself.

Please let me know if there is any problem. I'll try to solve it as soon as possible.
Thanks again :)

src/components/TooltipController/TooltipController.tsx
```diff
@@ -235,6 +237,24 @@ const TooltipController = ({
...
+    }
+    if (border) {
+      if (!cssAttrIsValid('border', border)) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `[react-tooltip] "${border}" is not a valid \`border\`. See https://developer.mozilla.org/en-US/docs/Web/CSS/border`,
+        )
+      }
+    }
+  }, [])
```
gabrieljablonski, line 256: @danielbarion see how you feel about validating props like this. Planning on doing the same for `opacity`.

gabrieljablonski (approved):

src/utils/css-attr-is-valid.ts
```diff
@@ -0,0 +1,25 @@
...
+  iframe.contentDocument.head.appendChild(style)
+  const { sheet } = style
+  if (!sheet) {
+    return true
+  }
+  const result = sheet.cssRules[0].cssText
+  iframe.remove()
+  const match = result.match(new RegExp(`${attr}:`))
+  return !!match
+}
```
danielbarion, line 25: What if instead of the iframe, we use CSS supports?

```
CSS.supports(propertyName, propertyValue)

CSS.supports('color','red') 
//True.
CSS.supports('color', '#007')
//True. 

CSS.supports('color', 'random')
//False. 
CSS.supports('colours', 'red')
//False. 
```

src/components/TooltipController/TooltipController.tsx
```diff
@@ -235,6 +237,22 @@ const TooltipController = ({
...
+      // eslint-disable-next-line no-console
+      console.warn('[react-tooltip] Do not set `style.border`. Use `border` prop instead.')
+    }
+    if (border && !cssAttrIsValid('border', border)) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `[react-tooltip] "${border}" is not a valid \`border\`. See https://developer.mozilla.org/en-US/docs/Web/CSS/border`,
+      )
+    }
+  }, [])
```
danielbarion, line 254: I have the feeling that this looks a bit verbose

danielbarion (changes requested):

src/components/TooltipController/TooltipController.tsx
danielbarion: looks a bit verbose to me, but I don't have a better idea at the moment

src/utils/css-attr-is-valid.ts
danielbarion: I'm not sure if this will work as we want, but it's just a suggestion

gabrieljablonski: This seems like exactly what I was trying to achieve. You can try on the browser console:

```js
> CSS.supports('border', '1px solid red')
< true
> CSS.supports('border', '1px solid redd')
< false
```

Will update later.

src/components/TooltipController/TooltipController.tsx
gabrieljablonski: We can drop the MDN docs link then.

danielbarion (approved):

PR #1052, opened 2023-07-02
feat: add event listener to handle style injection

danielbarion:
* add an event listener to handle when the styles should be injected

This also allows us we control the style injection by props

PR #1062, opened 2023-07-18
fix: add react_app prefix into the available env variables

### LINKED ISSUE #1060: [BUG] Unable to disable style injection
Hi,
I have to apply CSP standards in my React application. So, according to your docs, we have to set REACT_TOOLTIP_DISABLE_BASE_STYLES=true & REACT_TOOLTIP_DISABLE_CORE_STYLES=true to disable styles and import the CSS styles file in app.js. But apparently, after following all the steps I'm still seeing react-tooltip adding inline styles.

Can you please suggest a way how can I achieve this? You can check SS of the error.

image

Regards,
Shahwar

danielbarion:
close #1060


gabrieljablonski (approved):

danielbarion:
@gabrieljablonski please go ahead to merge it and do a new release when you have time, thanks!

gabrieljablonski:
@danielbarion I was actually still unable to make this work on a fresh CRA project. Looking into it.

gabrieljablonski:
@danielbarion Apparently using anything other than `process.env.NODE_ENV` from inside `node_modules` does not seem that trivial. Tried a few things and still couldn't get it to work.

Please also try looking into this whenever you have the time.

danielbarion:
@gabrieljablonski 
image
image

The issue is that `process` is `undefined` on the browser, even with CRA being able to get the envs

I'll try to solve it

danielbarion:
When I remove the validation:
image

image

gabrieljablonski:
Dropped env variables in favor of #1052

Issue #1057, opened 2023-07-11
[BUG] Tooltip is not tracking trigger element when resizing window


hurrmun:
**Bug description**
resizing the window does not move the tooltip along with the anchor element

**Version of Package**
v5.18.0

**To Reproduce**
codesandbox: **https://codesandbox.io/p/sandbox/react-tooltip-resize-tracking-mszsgp?file=%2Fsrc%2FApp.tsx%3A6%2C17** 

**Expected behavior**
Expected the tooltip to track the trigger element upon page resize

**Desktop (please complete the following information if possible or delete this section):**
 - OS: MacOS
 - Browser: Chrome
 - Frameworks [e.g. React 17/18, Vite]

**Additional context**
Not sure if this is considered a bug as the behaviour is not explicitly stated on the docs, but floating-ui seems to support the resize using autoUpdate and the library already tracks scrolling

gabrieljablonski:
We indeed do not track page resize. We'll leave this open so we can try using `autoUpdate()` in the future.

Until then, my suggestion is to use `closeOnResize`.

gabrieljablonski:
`react-tooltip@v5.20.0` now uses `floating-ui`'s `autoUpdate()` function.

PR #1075, opened 2023-07-30
Track window resize to reposition tooltip

Tanmayshetty:
Relevant issue: https://github.com/ReactTooltip/react-tooltip/issues/1057

### DIFF
--- a/docs/docs/troubleshooting.mdx
+++ b/docs/docs/troubleshooting.mdx
@@ -64,35 +64,6 @@ If you've imported the default styling and the tooltip is still not showing up w
 
 If `data-tooltip-content` and `data-tooltip-html` are both unset (or they have empty values) on the anchor element, and also the `content`, `render`, and `children` props on the tooltip are unset (or have empty values), the tooltip is not shown by default.
 
-
-## The tooltip doesn't move when scrolling
-
-If your anchor element is inside a scrolling element, your tooltip might get "stuck" in place when scrolling.
-There are two ways to avoid this.
-
-### Change your CSS (recommended)
-
-For the tooltip to be properly placed inside a scrolling element, the following conditions must be met:
-
-1. The tooltip component has to be inside the scrolling element (placing it as a direct child is **not** required)
-2. The `positionStrategy` tooltip prop must be unset, or set to the default (`absolute`)
-3. The scrolling element should have the CSS attribute `position: relative`
-
-:::info
-
-The `position: relative` attribute can be set on any element on the DOM structure between the scrolling element and the tooltip. 
-This means the tooltip component doesn't have to be a direct child of the scrolling element.
-
-:::
-
-### Use `closeOnScroll` prop
-
-```tsx
-<Tooltip closeOnScroll={true} />
-```
-
-When `closeOnScroll` is set, scrolling will immediately close the tooltip (`closeOnResize` also exists for closing when resizing the window).
-
 ## Bad performance
 
 If you're experiencing any kind of unexpected behavior or bad performance on your application when using ReactTooltip, here are a few things you can try.
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@types/css-modules": "^1.0.2",
     "@types/jest": "29.4.0",
     "@types/node": "^18.15.3",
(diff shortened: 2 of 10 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -290,6 +290,52 @@ const Tooltip = ({
   // mouse enter and focus events being triggered toggether
   const debouncedHandleShowTooltip = debounce(handleShowTooltip, 50, true)
   const debouncedHandleHideTooltip = debounce(handleHideTooltip, 50, true)
+  const updateTooltipPosition = () => {
```
Tanmayshetty, line 293: No changes were made to **updateToolTipPosition** method

gabrieljablonski:
Thanks for the contribution.

Have you seen the [`autoUpdate()` function](https://floating-ui.com/docs/autoUpdate) function? I believe we should use it.

For now, I suggest the options to be set `ancestorResize`, `elementResize`, and `layoutShift`.

Don't forget to call the cleanup function on the `useEffect()` return.

yarn.lock
Tanmayshetty, line 1: Updated **yarn.lock* file since the current version of **floating-ui/dom** did not support **layoutShift**

package.json
```diff
@@ -125,6 +124,7 @@
   },
   "dependencies": {
     "@floating-ui/dom": "^1.0.0",
+    "@types/react": "^18.2.17",
```
gabrieljablonski: Is this update really necessary? If not, please revert to `18.0.28` for consistency.

If it is, please let me know why, and also reinstall it as a dev dependency (`yarn add -D`)

gabrieljablonski:
I see you're still making some adjustments, but at first glance it looks ok.

Will review sometime this week.

Thanks again for the contribution!

package.json
```diff
@@ -59,7 +59,7 @@
     "@types/css-modules": "^1.0.2",
     "@types/jest": "29.4.0",
     "@types/node": "^18.15.3",
-    "@types/react": "18.0.28",
+    "@types/react": "^18.2.17",
```
Tanmayshetty, line 62: Updated this since I was getting the following warning messages during the build

> (!) Plugin typescript: @rollup/plugin-typescript TS2300: Duplicate identifier 'LibraryManagedAttributes'.
node_modules/@types/react-dom/node_modules/@types/react/ts5.0/index.d.ts: (3208:14)
...
...

gabrieljablonski:
We may need to mock `ResizeObserver` to get the tests to pass

https://stackoverflow.com/a/67006794

@Tanmayshetty no need to worry about this. I'll get to it when I have the time to fully review this later.

docs/docs/troubleshooting.mdx
```diff
@@ -64,35 +64,6 @@ If you've imported the default styling and the tooltip is still not showing up w
...
-:::
-
-### Use `closeOnScroll` prop
-
-```tsx
-<Tooltip closeOnScroll={true} />
-```
-
-When `closeOnScroll` is set, scrolling will immediately close the tooltip (`closeOnResize` also exists for closing when resizing the window).
-
```
gabrieljablonski, line 95: With `autoUpdate()` this should never be a problem anymore. Tested a bunch of different cases and all of them worked.

src/test/utils.spec.js
```diff
@@ -55,19 +55,7 @@ describe('compute positions', () => {
...
-        right: '',
-        top: '',
-      },
-      tooltipStyles: {
-        left: '5px',
-        top: '-10px',
-      },
-      place: 'top',
-    })
+    expect(value).toMatchSnapshot()
```
gabrieljablonski, line 58: Does this break anything? This makes it easier to maintain with `yarn test -u`.

gabrieljablonski (approved):

gabrieljablonski:
@Tanmayshetty made some adjustments, and everything's looking good!

Will wait for @danielbarion's review, then we'll merge and release this soon.

src/test/utils.spec.js
danielbarion: I believe this doesn't break anything,

thanks for the update!

danielbarion (approved):

PR #1106, opened 2023-10-31
Fade-out transition when closing the tooltip

### LINKED ISSUE #1105: [FEAT REQ] `react-tooltip__show` class counterpart for closing the tooltip
It is not currently possible to use a "closing transition" for the tooltip, similar to the fade in when it opens.
Adding a `react-tooltip__close` class (which requires some logic changes, since the tooltip currently closes immediately) should allow for using customizing the closing transition as well.

gabrieljablonski:
Closes #1105.

Reminder to link to https://react-tooltip.com/docs/examples/styling#customizing-openingclosing-animation on release notes.

Beta version `react-tooltip@5.22.0-beta.1106.0`

### DIFF
--- a/docs/docs/examples/styling.mdx
+++ b/docs/docs/examples/styling.mdx
@@ -381,7 +381,7 @@ If you wish to change the delay for any of them, override the following CSS vari
 
 Do not set `--rt-transition-closing-delay` to `0`. Doing so will result in the tooltip component being stuck (but not visible) on the DOM. This isn't itself a problem, but may lead to performance issues.
 
-Set to `1ms` or a similar value instead if you with to disable the fade-out transition when closing.
+Set to `1ms` or a similar value if you want to disable the fade-out transition when closing.
 
 :::
 
--- a/docs/docs/examples/styling.mdx
+++ b/docs/docs/examples/styling.mdx
@@ -372,6 +372,26 @@ In summary, if you do it correctly you can use CSS specificity instead of `!impo
 
 :::
 
+### Customizing opening/closing animation
+
+By default, the tooltip has a fade-in/fade-out transition when opening/closing, with a delay of 150ms for both.
+If you wish to change the delay for any of them, override the following CSS variables:
+
+:::caution
+
+Do not set `--rt-transition-closing-delay` to `0`. Doing so will result in the tooltip component being stuck (but not visible) on the DOM. This isn't itself a problem, but may lead to performance issues.
+
+Set to `1ms` or a similar value instead if you with to disable the fade-out transition when closing.
+
+:::
+
+```css
+:root {
+  --rt-transition-show-delay: 0.15s;
+  --rt-transition-closing-delay: 0.15s;
+}
+```
+
 ### Disabling ReactTooltip CSS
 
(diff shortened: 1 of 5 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -657,13 +639,21 @@ const Tooltip = ({
...
+      onTransitionEnd={(event: TransitionEvent) => {
+        /**
+         * @warning if `--rt-transition-closing-delay` is set to 0,
+         * the tooltip will be stuck (but not visible) on the DOM
+         */
+        if (show || event.propertyName !== 'opacity') {
+          return
+        }
+        setRendered(false)
+      }}
```
gabrieljablonski, line 656: Is there an obvious alternative to this? Thought about doing a check on the `--rt-transition-closing-delay` variable to setup a timeout for `setRendered(false)` (as it was being done before, see code removed above), but it felt very hacky.

src/tokens.css
```diff
@@ -6,4 +6,6 @@
   --rt-color-warning: #f0ad4e;
   --rt-color-info: #337ab7;
   --rt-opacity: 0.9;
+  --rt-transition-show-delay: 0.15s;
+  --rt-transition-closing-delay: 0.15s;
```
gabrieljablonski, line 10: After testing for some time, 0.3s started feeling waaaay too long. Still open for suggestions though.

gabrieljablonski:
@danielbarion Is it a bad idea to introduce the fade-out animation on a minor release? We have the option to "disable" it by setting the CSS variable to a low value, and have it like this by default only on v6.

It doesn't feel like a super huge change, but it's something to consider.

gabrieljablonski:
This also broke the `'tooltip with delay hide'` test. But since we already have other issues with testing, and our workflow is completely broken due to the issue with `iltorb` deprecation, I've chosen to ignore this for now.

docs/docs/examples/styling.mdx
```diff
@@ -372,6 +372,26 @@ In summary, if you do it correctly you can use CSS specificity instead of `!impo
...
+### Customizing opening/closing animation
+
+By default, the tooltip has a fade-in/fade-out transition when opening/closing, with a delay of 150ms for both.
+If you wish to change the delay for any of them, override the following CSS variables:
+
+:::caution
+
+Do not set `--rt-transition-closing-delay` to `0`. Doing so will result in the tooltip component being stuck (but not visible) on the DOM. This isn't itself a problem, but may lead to performance issues.
+
+Set to `1ms` or a similar value instead if you with to disable the fade-out transition when closing.
```
danielbarion, line 384: `if you want` maybe?

src/components/Tooltip/Tooltip.tsx
danielbarion: I do agree, we can find a better solution for this on V6

src/tokens.css
danielbarion: https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9

danielbarion (approved):

danielbarion:
@gabrieljablonski code looks good, just one typo on docs

danielbarion (approved):

PR #1109, opened 2023-10-31
Expose the tooltip ref to allow for imperative control

### LINKED ISSUE #1036: [FEAT REQ] setClose or close method
I know that there is an option to make the tooltip "hidden", but it does not help in my case, once we set "hidden" to "true", it does not appear anymore. In some situation, we need to manage the hide/show manually, for example I am using this package: https://github.com/JedWatson/react-select and I need to manage it on my side, if not Tooltip will not work.

What will be perfect, is to get a method like "close()" to be able to close the tooltip which will show again if we hover on the parent element. I saw that there is already a function named "handleShow", maybe you can just "export" it?

**Describe alternatives you've considered**

**Additional context**
If you make "handleShow" public, it should be fine, if not, adding "close()" and "show()" will be very useful.

gabrieljablonski:
Closes #1036.

- [x] Documentation
- [x] `imperativeModeOnly` prop after merge with open/close events PR (+ note on example)

Open for suggestions on other internal states to expose, or other methods to implement.

Beta version `react-tooltip@5.23.0-beta.1109.0`


danielbarion (approved):

danielbarion:
@gabrieljablonski there are some conflicts here, but the code looks good to me :)

PR #1117, opened 2023-11-12
Check if `CSS.supports()` is available

gabrieljablonski:


danielbarion (approved):

targumon (commented):
It's quite verbose. Personally I'd write it as:

```js
const cssSupports = (property: string, value: string): boolean => (
  'CSS' in window && 'supports' in window.CSS ? window.CSS.supports(property, value) : true
)
```
depending on your transpiler config maybe even:
```js
const cssSupports = (property: string, value: string): boolean => (
  window.CSS?.supports?.(property, value) ?? true
)
```

but LGTM!

src/utils/css-supports.ts
```diff
@@ -0,0 +1,9 @@
+const cssSupports = (property: string, value: string): boolean => {
```
targumon, line 1: It's quite verbose. Personally I'd write it as:

```js
const cssSupports = (property: string, value: string): boolean => (
  'CSS' in window && 'supports' in window.CSS ? window.CSS.supports(property, value) : true
)
```
depending on your transpiler config maybe even:
```js
const cssSupports = (property: string, value: string): boolean => (
  window.CSS?.supports?.(property, value) ?? true
)
```

but LGTM!

gabrieljablonski: Thanks for the suggestions.

The first one is not really in line with the style we use. A bit of verbosity for the sake of clarity is preferred.

For the second one, I definitely considered it, but since everything I could find about checking for browser features suggests using the `in` keyword, I'd rather do it like that. Though you're right, with our current transpiler options, pretty sure it would work.

Thanks again, there's a chance someday we'll look at it again and find it indeed too verbose 😅😅

src/utils/css-supports.ts
```diff
@@ -0,0 +1,6 @@
...
+  const hasCssSupports = 'CSS' in window && 'supports' in window.CSS
+  return hasCssSupports ? window.CSS.supports(property, value) : true
```
gabrieljablonski, line 3: There we go, should be a nice compromise 😀 @targumon

PR #1132, opened 2023-12-02
Support data-tooltip-class-name on anchor elements

fatton139:
As per the discussion here: https://github.com/ReactTooltip/react-tooltip/discussions/1129#discussioncomment-7717429

Support `data-tooltip-class-name` on anchor elements for injection into the tooltip container.

The tests doesn't look to be passing.


gabrieljablonski:
Thanks for taking the time, everything looks good on the first pass, will be testing and reviewing properly later.

And yes, our workflows are a bit broken right now, including testing, so don't worry about that.

gabrieljablonski (approved):
Everything looks good, no changes needed.

As for the test, we're already planning on reviewing all test cases to fix some inconsistencies, so no need to worry about it passing right now.

danielbarion (approved):

danielbarion:
@fatton139 we are still working on a roadmap for the V6 and we don't have a release date for it yet, I believe this PR is good to go on V5, thanks for checking!

gabrieljablonski:
> Should I be targeting the v6 branch?

No need, this will me merged to master soon.

PR #1153, opened 2024-01-18
Fix: remove anchor when `data-tooltip-id` gets changed

mixrich:
When `data-tooltip-id` gets updated from `A` to `B` for the same DOM element, the `A` still continues showing the tooltip when it isn't expected. The PR fixes that, now `A` will properly untrack the tooltip when `data-tooltip-id` becomes not `A`

UPD: I see some failed tests. Looks like one test uses the outdated snapshot, the second one seems to be flaky. The same happens in `master` as well

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -643,6 +643,9 @@ const Tooltip = ({
           const newId = (mutation.target as HTMLElement).getAttribute('data-tooltip-id')
           if (newId === id) {
             newAnchors.push(mutation.target as HTMLElement)
+          } else if (mutation.oldValue === id) {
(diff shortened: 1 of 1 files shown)


gabrieljablonski:
Thanks for your contribution! 

Some quick testing shows this works as expected. After @danielbarion also approves it, we'll merge a do a release.

---

> UPD: I see some failed tests. Looks like one test uses the outdated snapshot, the second one seems to be flaky. The same happens in master as well

Yep, we've got some open issues with our test specs, don't worry about those.

gabrieljablonski (approved):

danielbarion (approved):

mixrich:
Awesome, thank you!

PR #1168, opened 2024-02-10
Improve `delayShow`/`delayHide` behavior

### LINKED ISSUE #1119: [BUG] Unexpected behavior with `delayShow`
**Bug description**
`delayShow={1000}` results in unexpected behavior when hovering between anchor elements.

**Version of Package**
v5.22.0

**To Reproduce**
Hover between two anchor elements that share a tooltip with a value set for `delayShow`. The tooltip will change to the second anchor element before closing.

**Expected behavior**
Two options:

1. The tooltip should stay open when hovering between two elements that are close enough
2. The tooltip should abruptly close, and open again for the new anchor only after the set delay

I lean towards 1.

**Additional context**
CodeSandbox sample: https://codesandbox.io/s/frosty-noether-y4m79q?file=/src/App.js

#1114 related

### LINKED ISSUE #1162: [BUG] Tooltip shows briefly when using delayShow and moving to directly adjacent element
It seems that when using `delayShow`  and moving from one tooltip element to another that's immediately next to it, the "new" tooltip for the element you just moved to flashes for a second, hides, and then shows again after the time delay. Removing your mouse entirely and trying again works fine. 

So it seems to be related to moving from one tooltip element to another.

```tsx
    <Tooltip
      className={styles.tooltip}
      delayShow={1000}
      id='default'
      noArrow
      place='top-start'
    />
    
<span
  data-tooltip-content='This is a tooltip'
  data-tooltip-id='default'
  data-tooltip-variant='info'>test</span>
<span
  data-tooltip-content='This is a tooltip'
  data-tooltip-id='default'
  data-tooltip-variant='info'>test</span>
```

gabrieljablonski:
Closes #1119.
Closes #1162.

Beta version `react-tooltip@5.26.1-beta.1168.0`

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -635,6 +635,8 @@ const Tooltip = ({
     closeEvents,
     globalCloseEvents,
     shouldOpenOnClick,
+    delayShow,
+    delayHide,
   ])
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


gabrieljablonski:
@danielbarion you can give it a quick look if you want, but I still haven't tested if this introduces any issues.

danielbarion (approved):

PR #1171, opened 2024-02-10
Fix testing and minor refactorings

### LINKED ISSUE #1080: [BUG] Test `'tooltip with delay show'` inconsistent snapshot
Managed to reproduce locally the problem with the workflow sometimes failing, and passing on re-run, though this time it is happening with the newer `react-tooltip__show` class.

https://github.com/ReactTooltip/react-tooltip/blob/master/src/test/tooltip-props.spec.js#L118-L144

https://github.com/ReactTooltip/react-tooltip/blob/master/src/test/__snapshots__/tooltip-props.spec.js.snap#L86

The snapshot switches between having and not having `react-tooltip__show`. I believe this is likely because we're not using `waitFor()` correctly to account for the `delayShow` timer.

gabrieljablonski:
Closes #1080.

Might close other test related issues. Will have to check.


src/utils/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type { IComputePositions }
+export {
+  computeTooltipPosition,
+  cssSupports,
+  cssTimeToMs,
+  debounce,
+  getScrollParent,
+  useIsomorphicLayoutEffect,
+}
```
gabrieljablonski, line 17: We had some inconsistent exports here. All util functions are now default exported, and available with named export from `utils/index.ts`.

gabrieljablonski:
For future reference, when testing the tooltip **NOT** to be on screen, we recommend the following (using `@testing-library/react`):

```js
await waitFor(() => {
  expect(screen.queryByRole('tooltip')).not.toBeInTheDocument()
})
```

When getting a reference to the tooltip, we recommend this:

```js
const tooltip = await screen.findByRole('tooltip')
expect(tooltip).toHaveAttribute('style')
```

Waiting for the tooltip to have the `react-tooltip__show` (or `react-tooltip__closing`) class is often not necessary, but may be useful when testing both opening and closing the tooltip on the same spec.

```js
await waitFor(() => {
  // also useful with `react-tooltip__closing`
  expect(screen.queryByRole('tooltip')).toHaveClass('react-tooltip__show')
})
```

danielbarion (approved):

PR #1179, opened 2024-03-14
V6

### LINKED ISSUE #1010: [BUG] Tooltip position and show status not synchronized.
**Describe the bug**
When moving from one anchor to another, sometimes the tooltip's position is updated before its visibility. This results in some glitchy / flickering behavior.

**Version of Package**
v5.10.6

**To Reproduce**
1. Set `Tooltip` `delayShow` to `300`
2. Set `Tooltip` `delayHide` to `0`
3. Move from one anchor to another (with no space between)
4. Observe that the tooltip first changes position and content, *then* hides. Then after 300ms shows again.

**Expected behavior**
The tooltip should hide *first*, then update position and content.

**Screenshots**
https://user-images.githubusercontent.com/431251/230918927-995f1ccf-2449-40d8-8a4b-948c48605393.mp4

 - OS: MacOS
 - Browser firefox
 - Frameworks `create-react-app`

**Additional context**
Repro: https://github.com/knubie/react-tooltip-demo
Somewhat, but not really related: https://github.com/ReactTooltip/react-tooltip/issues/1008

### LINKED ISSUE #1074: Update issue templates to use issue forms
Issue forms are pretty great, we should definitely use them.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

### LINKED ISSUE #1163: [FEAT REQ] Render tooltips inside portal
Rendering tooltips inside a portal would resolve issues caused position absolute/fixed. Fixed works in most cases, but some libraries like Storybook use `transform/scale` on their story containers which counts as a new "container" for `position: fixed` elements. This causes the tooltip to render incorrectly in auto-generated storybook documents because the x/y are based in the viewport, but the container is no longer the viewport. The tooltips appear at incorrect locations and often cause overflow/scrolls in storybook.

Rendering in portals is a common solution to this, our context/dropdown menu and select input libraries use this for rendering the menus.

### LINKED ISSUE #1212: Is there a tooltip duration change function?
When mouse hover, the tooltip is displayed, and I want to close it 5 seconds after the tooltip is displayed.

I customized the tooltip using afterShow and afterHide, but there is a problem.

If you hover your mouse over id=test1 and move your mouse to id=test2 to open the tooltip and mouse over the existing test1 again, there is a bug that closes immediately, not after 5 seconds.

When the id changes, I want to initialize the timeOut of that id when I do mouseLeave.

I'm going to solve this method and set the tooltip duration to 5 seconds, so if there's a good way other than this, please share it.

- condeSandBox
https://codesandbox.io/p/sandbox/tooltip-w2q7ll?file=%2Fsrc%2FApp.tsx%3A15%2C14-15%2C19

danielbarion:
- [x] cleanup of the deprecated options of the Tooltip
- [x] explore new bundler options ([bun.sh](https://bun.sh)?)
- [x] Rollup upgraded v2 -> v4

Note: we need to consider the custom rollup plugin we did to inject the CSS into the build file (`/rollup-plugins/replace-before-save-file.js`)
- [x] dismiss and/or fix [dependabot security alerts](https://github.com/ReactTooltip/react-tooltip/security/dependabot)
- [x] improve tooltip bundle size
* check global average browser version and reduce the range on the build that includes polyfills ✅
* we have different ways to use the tooltip. Check [code splitting](https://rollupjs.org/tutorial/#code-splitting) to reduce the loaded code related to the way the project wants to use react-tooltip
- [x] improve tooltip performance - how to benchmark it to compare with v5?
- [x] improve internal state handling (internal state should be updated in the least render cycles possible)
- [x] review all `useMemo()`/`useCallback()`/`useEffect()` dependencies (the `react-hooks/exhaustive-deps` rule should only be ignore if strictly necessary, i.e. an specific dependency breaks functionality)
- [x] break down large effects and functions into more concise (and properly named) functions
- [x] review docs to make sure it's up to date with V6 changes after PR gets ready to be reviewed
- [x] write the tests to cover the library code and his utils

Fixes #1163
Fixes #1212
Fixes #1074
Fixes #1010

Related discussion: https://github.com/ReactTooltip/react-tooltip/discussions/1253

## Summary by CodeRabbit

* **New Features**
  * Added a Benchmark page and CLI/browser benchmark tools with snapshot results.

* **Breaking Changes**
  * Removed raw-HTML tooltip APIs (`data-tooltip-html`, `html` prop); use children or render for rich content.
  * Removed legacy provider/wrapper and several deprecated props/events; simplified tooltip API surface.

* **Documentation**
  * Major docs refresh: versioned v5 docs, V5→V6 upgrade guide, expanded examples and troubleshooting.

* **Chores**
  * Package bumped to v6.0.0 with modernized build/test tooling.


danielbarion:
* Bun is amazing but unfortunately not an option at the moment because of:

[image
](https://bun.sh/docs/bundler)

danielbarion:
image

danielbarion:
> * Bun is amazing but unfortunately not an option at the moment because of:
> 
> [image ](https://bun.sh/docs/bundler)
> 
> Also, note added:
> 
> ` Note: we need to consider the custom rollup plugin we did to inject the CSS into the build file (/rollup-plugins/replace-before-save-file.js)`

Almost there:

image

danielbarion:
Switched from Draft to trigger the beta release.

PR #1194, opened 2024-06-06
Add `mouseover` and `mouseout` as default events

### LINKED ISSUE #1193: Nested elements with tooltip
Hi!

I have nested elements with tooltips

<div data-tip={tooltipContent1}>
    <div data-tip={tooltipContent2}/>
</div>

but when I move cursor from child DOM element to parent, tooltip data doesn't change

this is how it looks: https://prnt.sc/10h83bt

reproducible demo: https://codesandbox.io/p/sandbox/react-tooltip-issue-y867xh?file=%2Fsrc%2FApp.js%3A6%2C32

is there a way to solve this?

gabrieljablonski:
Closes #1193 

See [this demo](https://www.w3schools.com/jquery/tryit.asp?filename=tryjquery_event_mouseenter_mouseover) for explanation. `mouseover` and `mouseout` apply to a wider range of scenarios, more specifically when dealing with nested tooltip anchors.

`mouseenter` and `mouseleave` should still be present for backwards compatibility. But should we deprecate it and remove it on v6? Keeping it seems fine for more functionality, but will likely almost never be used.

This will not impact current behavior for "regular" tooltip usage (no nested anchor elements).

## Summary by CodeRabbit

- **New Features**
  - Enhanced Tooltip component with new event handling options including `mouseover` and `mouseout`.

- **Bug Fixes**
  - Improved precision in managing tooltip open and close events.

- **Documentation**
  - Updated Tooltip component documentation to reflect new event handling options and deprecated props.


src/components/Tooltip/TooltipTypes.d.ts
```diff
@@ -88,13 +88,15 @@ export interface TooltipRefProps {
 export type AnchorOpenEvents = {
   mouseenter?: boolean
```
gabrieljablonski, line 89: Should we deprecate `mouseenter` and `mouseleave`?

Keeping them seems fine, but with `mouseover`/`mouseout`, would they ever really be useful?

danielbarion: What will be the benefits to keep and to remove?

gabrieljablonski: I honestly can't think of any realistic situation `mouseenter`/`mouseleave` (1) would solve something that `mouseover`/`mouseout` (2) doesn't.

The only issue with (2) is that we're triggering sequential `hide`/`show` state updates on the tooltip every time the user hovers between the nested element and the parent element. I couldn't visually notice any bad behavior due to this, but it's something to keep in mind.

Maybe we just leave (1) as the default, and have (2) + a troubleshooting page for people having issues with nested anchor elements?

danielbarion: Got it, about removing it: if it's not causing any bug, I would like to keep it so we support those events.

About the `contains()`, I like the idea

danielbarion (approved):

PR #1202, opened 2024-06-26
Update TooltipControllerTypes.d.ts

devyanshyadav:
Allow JSX.Element in tooltip content and html props

- Update ITooltipController interface
- content and html props now accept string | JSX.Element
- Improves flexibility for adding React elements in TypeScript

## Summary by CodeRabbit

- **New Features**
  - Tooltip content can now include JSX elements, allowing for richer and more dynamic tooltip displays.

### DIFF
--- a/src/components/TooltipController/TooltipControllerTypes.d.ts
+++ b/src/components/TooltipController/TooltipControllerTypes.d.ts
@@ -17,11 +17,11 @@ import type {
 export interface ITooltipController {
   className?: string
   classNameArrow?: string
-  content?: string
+  content?: string | JSX.Element
   /**
    * @deprecated Use `children` or `render` instead
    */
-  html?: string
+  html?: string | JSX.Element
   render?: (render: { content: string | null; activeAnchor: HTMLElement | null }) => ChildrenType
   place?: PlacesType
   offset?: number


gabrieljablonski:
I'll have to check if this doesn't break anything (we might assume `content` is always a `string` somewhere), but at first glance it technically doesn't add anything new to the tooltip (just some convenience, I guess), since you can already pass `JSX` as children to the tooltip, or with the `render` prop. For instance:

```tsx
<Tooltip>
  <MyTooltipContent />
</Tooltip>
```

We're already doing some planning for V6 of the tooltip, so we might integrate this, or just drop the `content` prop entirely in favor of `children`, so there's less redundancy in existing features (`html` is already deprecated).

Not sure yet what we'll do, but thanks for the suggestion!

gabrieljablonski:
Sorry for taking so long to get back to this.

Did some quick testing and this change breaks a couple of type references across the codebase, so it would require some more refactoring. Although it wouldn't be too hard, I don't think it is worth it. In the v6 we might even drop the `content` prop (`html` is already deprecated), since it is redundant with using `children`.

If you want to pass a JSX element to `content` prop, just pass it as the child element for the tooltip component, as suggested in the previous comment.

```tsx
<Tooltip>
  <MyTooltipContent />
</Tooltip>
```

PR #1203, opened 2024-07-02
Remove warning about styles already being injected

### LINKED ISSUE #1185: [BUG] Element with id 'react-tooltip-base-styles' already exists
### Discussed in https://github.com/ReactTooltip/react-tooltip/discussions/1184

Check the discussion for more info.

gabrieljablonski:
Closes #1185.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of elements with duplicate IDs to prevent potential conflicts when the tooltip is imported by multiple modules.

### DIFF
--- a/src/utils/handle-style.ts
+++ b/src/utils/handle-style.ts
@@ -52,13 +52,7 @@ function injectStyle({
   const { insertAt } = ref
 
   if (document.getElementById(id)) {
-    // this should never happen because of `injected[type]`
-    if (process.env.NODE_ENV !== 'production') {
-      // eslint-disable-next-line no-console
-      console.warn(
-        `[react-tooltip] Element with id '${id}' already exists. Call \`removeStyle()\` first`,
-      )
-    }
+    // this could happen in cases the tooltip is imported by multiple js modules
     return
   }
 


src/utils/handle-style.ts
```diff
@@ -52,13 +52,7 @@ function injectStyle({
...
 
   if (document.getElementById(id)) {
-    // this should never happen because of `injected[type]`
-    if (process.env.NODE_ENV !== 'production') {
-      // eslint-disable-next-line no-console
-      console.warn(
-        `[react-tooltip] Element with id '${id}' already exists. Call \`removeStyle()\` first`,
-      )
-    }
+    // this could happen in cases the tooltip is imported by multiple js modules
```
gabrieljablonski, line 55: Is this even worth investigating further?

There seems to be no downside to just skipping injecting styles if they're already present.

danielbarion: I do agree, the issue is inject the styles multiple times, but there's no issue if we only ignore it if it's already injected into the website

danielbarion (approved):

Issue #1210, opened 2024-07-20
[FEAT REQ] Dynamic callback for disabling tooltip


nareshpingale:
In my application, I work with large datasets displayed in a data grid, where each cell has a tooltip. This extensive use of tooltips results in significant scripting time whenever a user interacts with them. The addEventListener and removeEventListener functions run on every cell and tooltip across the app. Consequently, even simple actions like scrolling or hovering through different cells through the table cause a considerable amount of scripting activity, impacting performance.

Currently, I've implemented a workaround tailored to my use case. Tooltips are necessary only when the data within a cell is clipped and an ellipsis appears. However, this solution is limited and not ideal for broader scenarios as it helps only because it reduces the amount of tooltips instead of solving the core problem.

I propose adding a `data-tooltip-dynamic-hide` attribute that accepts a callback function.

This callback would:
- Receive the anchor element and anything else that you think would be ideal exposing outside to the users from inside of Tooltip.tsx
- User would add logic using this callback. In my case compares the scrollWidth and offsetWidth properties to determine if the content is clipped.
`
data-tooltip-dynamic-hide={(anchorElement)=>anchorElement.scrollWidth <= anchorElement.offsetWidth}
`
- Decide whether to add the element as an anchor for the tooltip based on the result inside Tooltip.tsx

This approach would allow for more flexible and efficient tooltip management, accommodating various use cases and improving overall performance.

**Alternatives I have explored for far**
- I have explored using data-tooltip-hidden attribute. But access correct ref and in such cases inside .map become troublesome.
- As temporary workaround I have added show delays as to act as some kind of throttle.
- Considered a broader fix adding throttle over useEffect callback which adds and removes listeners. 

**Additional context**
I will be happy to open a PR introducing this feature or discussing other solutions around this.

Tooltips added on most basic example of tanstack table.
https://codesandbox.io/p/sandbox/angry-night-32nr8q

<img width="1267" alt="image" src="https://github.com/user-attachments/assets/b5a9d6d5-345c-4941-8855-74e7119d2e55">

gabrieljablonski:
I'll take a deeper look when I got some more time, but IIRC, your suggestion is not exactly viable. I believe a `data-*` can only really be  string (see [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes)).

[...]

PR #1211, opened 2024-07-21
feat: support for disabled tooltip callback

nareshpingale:
Addresse #1210 

Usage:
```
   <button
      data-tooltip-id="anchor-select"
      data-tooltip-content="this content is different"
      data-tooltip-dynamic="true"
      style={{
        maxWidth: '200px',
        overflow: 'hidden',
        textOverflow: 'ellipsis',
        whiteSpace: 'nowrap',
      }}
    >
      Anchor select with dynamic tooltip and a very long label to test overflow
    </button>
    <button data-tooltip-id="anchor-select" data-tooltip-dynamic="true">
      Anchor select with dynamic tooltip
    </button>
    <button data-tooltip-id="anchor-select">Anchor select without dynamic tooltip</button>
  </p>
  <Tooltip
    id="anchor-select"
    disableTooltip={(anchor) => {
      const { scrollWidth, offsetWidth } = anchor as HTMLElement
      if (anchor?.getAttribute('data-tooltip-dynamic') && scrollWidth <= offsetWidth) {
        return true
      }

      return false
    }}
  >
    Tooltip content
  </Tooltip>
```

https://github.com/user-attachments/assets/932b76f2-8733-465a-bf02-8970da36435c

## Summary by CodeRabbit

- **New Features**
	- Introduced a new optional prop `disableTooltip` for the Tooltip component, allowing developers to conditionally disable tooltips for specific anchors.
	- Enhanced the TooltipController component to support the new `disableTooltip` functionality, improving control over tooltip display behavior.
  
- **Documentation**
	- Updated documentation to include the new `disableTooltip` function, providing guidance on controlling tooltip visibility based on anchor states.
  
- **Tests**
	- Added test cases to validate the behavior of the `disableTooltip` prop, ensuring tooltips render correctly based on the prop's return value.


nareshpingale:
@gabrieljablonski do let me know if we need to update /examples for this. Also if everything looks fine at first glance will start adding changes to /docs.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -459,6 +460,9 @@ const Tooltip = ({
     const elementRefs = new Set(anchorRefs)
 
     anchorsBySelect.forEach((anchor) => {
+      if(disableTooltip?.(anchor)){
+        return;
+      }
```
gabrieljablonski, line 465: This separation between `anchorSelect` and `anchorId` is something we're planning to improve on for v6 (`anchorId` is already deprecated), but until then, we should probably do the check below as well.

Maybe like this:

```ts
    const anchorById = document.querySelector<HTMLElement>(`[id='${anchorId}']`)
    if (anchorById && !disableTooltip?.(anchorById)) {
      elementRefs.add({ current: anchorById })
    }
```

gabrieljablonski (commented):
Looks good for a first pass. If you don't mind, please add it to the "options" table on the docs.

Also, if you could try adding a test for the new prop under this [tests file](https://github.com/ReactTooltip/react-tooltip/blob/master/src/test/tooltip-props.spec.js), we'd appreciate it. No need to waste too much time on it though if you can't figure it out.

When those are done we'll do a last check again and see if we catch anything. The feature is simple enough, but just to make sure.

src/components/Tooltip/Tooltip.tsx
danielbarion: Please add one space between `if` and `(` on line 463

nareshpingale:
- Added tests for disableTooltip returning true and false
- Addressed formatting issue raised by @danielbarion 
- Added disableTooltip check to anchorById
- Added disableTooltip in options in docs, although wasn't able to verify this as there is a peer dependency error by our docusaurus version for react 18

gabrieljablonski (approved):
LGTM. Merging soon.

danielbarion (approved):

PR #1236, opened 2024-11-27
Added enforceDelayShow prop. (Fixes #1235)

### LINKED ISSUE #1235: [BUG] When using the same single tooltip on a page, the delayShow/data-tooltip-delay-show is not respected
**Bug description**
When using the same single tooltip on a page, the delayShow/data-tooltip-delay-show is not respected if you hover over them immediately after each other.

**Version of Package**
v5.28.0

**To Reproduce**
[Steps to reproduce the behavior.](https://stackblitz.com/edit/stackblitz-starters-apjjpf?file=app%2Fpage.tsx)

steps to reproduce in the attached stackblitz:

1.  hover over "A tooltip" for 1 second
2.  the tooltip is shown
3.  immediately hover over "Another tooltip"

actual result: the "Another tooltip" is shown immediately
expected result: the tooltip should take 1 second before it is shown

**Screenshots**

https://github.com/user-attachments/assets/3a6f0157-4702-4f98-8546-3fa91b7e18c9

 - OS: Windows 11
 - Browser: Microsoft Edge
 - Version: 131.0.2903.51 
 - Frameworks: React 18.2.0

Ville-Wennerlund:
This will make the delayShow prop to always be enforced even if the tooltip instance is rendered.

This can be used for when you only have one tooltip instance, but multiple tooltip anchors, and want it to always take the delayShow amount of time before it is shown.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new prop `enforceDelayShow` for the `Tooltip` component, allowing users to control the display timing of tooltips with more precision.
- **Improvements**
	- Enhanced tooltip behavior to respect display delays even when previously rendered, providing a more consistent user experience.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -68,6 +68,7 @@ const Tooltip = ({
   border,
   opacity,
   arrowColor,
+  enforceDelayShow = false,
   role = 'tooltip',
 }: ITooltip) => {
   const tooltipRef = useRef<HTMLElement>(null)
@@ -259,7 +260,7 @@ const Tooltip = ({
   const handleShowTooltipDelayed = (delay = delayShow) => {
     clearTimeoutRef(tooltipShowDelayTimerRef)
 
-    if (rendered) {
+    if (rendered && !enforceDelayShow) {
       // if the tooltip is already rendered, ignore delay
       handleShow(true)
       return
--- a/src/components/Tooltip/TooltipTypes.d.ts
+++ b/src/components/Tooltip/TooltipTypes.d.ts
@@ -158,5 +158,6 @@ export interface ITooltip {
   border?: CSSProperties['border']
   opacity?: CSSProperties['opacity']
   arrowColor?: CSSProperties['backgroundColor']
+  enforceDelayShow?: boolean
   role?: React.AriaRole
 }


Ville-Wennerlund:
I went back to version 5.24.0.
Essentially before the skip if rendered was added. Tooltip.jsx row 260ish.

And I assume I would get the same issues with my changes that I am proposing. That the tooltip blinks between the old anchor value and the new anchor value.

gabrieljablonski:
> I went back to version 5.24.0.

Tried approving the beta release so you could use it in your package until we have time to review this, but it didn't work. But since 5.24.0 apparently solved it for you, it's fine for now then.

Thanks for the contribution, will go over it when I have the time.

gabrieljablonski:
@Ville-Wennerlund sorry haven't had much time to go over stuff here.

Your change is simple enough, that I don't think it should break anything.

But please update the options table on the docs, and I'll do a quick test later and possibly do a release with the new option.

Ville-Wennerlund:
> I think you misunderstood me.
> 
> What I'm trying to say is that 5.24.0 and my proposed changes will probably behave the same.
> 
> 5.24.0 has the issue that if you hover between anchors that have the same tooltip instance, the tooltip will first disappear and reappear on the new anchor.
> 
> I will see if I can find a related issue to link to you, since I hope that someone else can explain it better than me : )
> 
> Edit: this issue: #1162 and your changes that fixed that issue: https://github.com/ReactTooltip/react-tooltip/pull/1168/files

I'm referring to this comment. 
I understand that you do not have the time, we will have to do some kind of work around for the time being.

